PR #3, opened 2020-06-12
Define ExtendedPlatform to represent a platform with extra features

giordano:
This is my initial attempt to define a new type which wraps a standard platform and extends it with some extra features

Still need to implement:
* [x] `Pkg.BinaryPlatform.platforms_match`


src/Platforms.jl
```diff
@@ -14,3 +16,201 @@ Pkg.BinaryPlatforms.triplet(::AnyPlatform) = "any"
...
+    print(io, BinaryPlatforms.platform_name(ep), "(", ep.p)
+    length(ep.ext) > 0 && print(io, ";")
+    join(io, [" $(k)=$(repr(v))" for (k, v) in ep.ext], ",")
+    print(io, ")")
+end
+
+# Vendor Pkg.BinaryPlatforms.platform_key_abi, with a different name.  The
+# difference with the original version is that if the triplet can't be parse we
+# return `nothing`, wthout issuing a warning
+function tryparse_standard_platform(machine::AbstractString)
```
staticfloat, line 51: It's up to you whether the performance hit of disabling logging is worth it, but I am always loathe to duplicate code, since I've already committed that sin many times in my life.  ;)

```
 using Logging, Pkg.BinaryPlatforms
function tryparse_standard_platform(machine::AbstractString)
    p = with_logger(NullLogger()) do
        BinaryPlatforms.platform_key_abi("dsgaeg")
    end
    if isa(p, UnknownPlatform)
        return nothing
    end
    return p
end
```

src/Platforms.jl
```diff
@@ -14,3 +16,201 @@ Pkg.BinaryPlatforms.triplet(::AnyPlatform) = "any"
...
+    extra_str = String[]
+    base_platform = platform_key_abi()
+    # Try to parse the longest string possible with the standard parser
+    while length(split_str) > 0
+        base_platform = tryparse_standard_platform(join(split_str, '-'))
+        if base_platform !== nothing
+            break
+        end
+        push!(extra_str, pop!(split_str))
+    end
```
staticfloat, line 172: I wonder if we can take the split pieces, then just filter out anything that doesn't have a `+` in it, then past those pieces to `tryparse_standard_platform`?  Anything that is "extended" should have `+` in it somewhere, and I don't think `+` can legally appear in a standard platform.

src/Platforms.jl
```diff
@@ -14,3 +16,201 @@ Pkg.BinaryPlatforms.triplet(::AnyPlatform) = "any"
...
+# same keys in the dictionaries have the same values
+Pkg.BinaryPlatforms.platforms_match(a::ExtendedPlatform{P}, b::ExtendedPlatform{P}) where {P<:Platform} =
+    platforms_match(a.p, b.p) &&  all(a.ext[k] == b.ext[k] for k in keys(a.ext) ∩ keys(b.ext))
+# The wrapped platform is the same as the other one: the arguments match if the
+# wrapper platform and the other one match
+Pkg.BinaryPlatforms.platforms_match(a::ExtendedPlatform{P}, b::P) where {P<:Platform} = platforms_match(a.p, b)
+Pkg.BinaryPlatforms.platforms_match(a::P, b::ExtendedPlatform{P}) where {P<:Platform} = platforms_match(a, b.p)
+# Extending a platform different from the other one: do not match
+Pkg.BinaryPlatforms.platforms_match(a::ExtendedPlatform{P1}, b::P2) where {P1<:Platform, P2<:Platform} = false
+Pkg.BinaryPlatforms.platforms_match(a::P1, b::ExtendedPlatform{P2}) where {P1<:Platform, P2<:Platform} = false
```
staticfloat, line 213: Oooh, that's a nice, tricksy case.  Good job.

test/platforms.jl
```diff
@@ -18,6 +18,117 @@ using BinaryBuilderBase: abi_agnostic
                    [Linux(:i686, libc=:glibc), Linux(:i686, libc=:musl), MacOS(:x86_64), Windows(:x86_64)]
 end
 
+@testset "ExtendedPlatform" begin
+    p = ExtendedPlatform(Linux(:x86_64; libc=:musl); microarchitecture = :avx, cuda = "9.2")
+    @test p.p == Linux(:x86_64, libc=:musl)
+    @test p.ext == Dict("microarchitecture" => "avx", "cuda" => "9.2")
```
staticfloat, line 24: Just for the record, I think I'm going to want the key to be `march` instead of `microarchitecture` just because that's such a mouthful.  Doesn't matter here though.  :P

src/Platforms.jl
```diff
@@ -14,3 +16,121 @@ Pkg.BinaryPlatforms.triplet(::AnyPlatform) = "any"
 # Fallback on x86_64-linux-musl, but this shouldn't really matter in practice.
 Pkg.BinaryPlatforms.arch(::AnyPlatform) = arch(Linux(:x86_64, libc=:musl))
 Base.show(io::IO, ::AnyPlatform) = print(io, "AnyPlatform()")
+
+## The extended platforms, to represent platforms beyond the standard ones in
+## Pkg.BinaryPlatforms.
```
staticfloat, line 21: If necessary, integrate this information into the docstring.

src/Platforms.jl
```diff
@@ -14,3 +16,121 @@ Pkg.BinaryPlatforms.triplet(::AnyPlatform) = "any"
...
+"""
+    ExtendedPlatform(p::Platform; kwargs...)
+
+Extend the given platform with extra information.  The key and the value can be
+arbitrary, with the conditions that they should only use alhanumerical
+characters, the underscore, or the dot (the latter only for the value).
```
staticfloat, line 27:
```suggestion
Extend a `Pkg.BinaryPlatforms.Platform` object with extra key-value mappings.
Arbitrary `String` keys and values are supported, with the constraint that
all strings should only use alphanumeric characters, the underscore, or the
dot (the latter only for the value).
```

PR #8, opened 2020-06-26
Use `-mtune=generic` for `x84-64` microarchitecture

giordano:


staticfloat:
`x84-64` 😂

giordano:
Look at the branch name :sweat_smile:

staticfloat:
`x64-64` == `x`? ;)

StefanKarpinski:
Your math is a bit rust, @staticfloat, that should be `(x-1)64`.

staticfloat:
And here I never thought I’d need to do algebra again after leaving school....

PR #10, opened 2020-06-28
Add `extended_platform_key_abi`

giordano:
Alternative to #6.  This one is based on libLLVM and doesn't require extra packages

### DIFF
--- a/src/Platforms.jl
+++ b/src/Platforms.jl
@@ -1,6 +1,6 @@
 using Pkg.Artifacts, Pkg.BinaryPlatforms, Logging
 
-export AnyPlatform, ExtendedPlatform, base_platform
+export AnyPlatform, ExtendedPlatform, base_platform, extended_platform_key_abi
 
 """
     AnyPlatform()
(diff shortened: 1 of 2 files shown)


src/Platforms.jl
```diff
@@ -207,3 +207,55 @@ function Pkg.Artifacts.pack_platform!(meta::Dict, p::ExtendedPlatform)
     # override the default ones
     Artifacts.pack_platform!(meta, base_platform(p))
 end
+
+# Get the list of CPU features for the host system querying libLLVM
+function get_cpu_features()
+    if VERSION < v"1.4"
+        error("Cannot automatically detect the features of the CPU with this version of Julia!")
+    end
```
giordano, line 215: If we go by this route, all the microarchitecture stuff will work only with Julia v1.4+, as this requires libllvm v8+.  Not too bad since the alternative would be to have some code in Base to do the same, but that would require Julia v1.6 or more.

src/Platforms.jl
```diff
@@ -207,3 +207,55 @@ function Pkg.Artifacts.pack_platform!(meta::Dict, p::ExtendedPlatform)
...
+                    return "avx512"
+                end
+                return "avx2"
+            end
+            return "avx"
+        end
+        return "x86_64"
+    else
+        @warn "Cannot determine the microarchitecture for the given set of features!"
+        return nothing
```
giordano, line 240: I don't want a hard failure if we can't determine the microarchitecture from the given set of features, but it's better to issue a warning just in case we're we're getting strange results.  Also, this function would need to be used only when we're asking the set of features for an architecture we know we support (currently only `x86_64`), as done below in `extended_platform_key_abi`

giordano:
Trying to collect what `LLVMGetHostCPUFeatures` returns on some aarch64 machines.  AWS Graviton 2 (should have [Neoverse N1](https://en.wikipedia.org/wiki/ARM_Cortex-A76#Neoverse_N1) chips), Travis CI (according to https://docs.travis-ci.com/user/reference/overview/ "LXD compliant OS images for arm64 are run in Packet", no more details), and Cloud Drone (which according to https://blog.drone.io/drone-cloud/ and https://www.packet.com/cloud/servers/c1-large-arm/ use Cavium ThunderX CPUs, without further details) all give:
```julia
["+neon", "+fp-armv8", "+crypto", "+crc"]
```
Thanks to @ianshmean for the collaboration!

giordano:
Superseded by #38

PR #13, opened 2020-07-09
Do not follow symlinks by default for `DirectorySource`

giordano:
Fix https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/845.  CC: @Keno


Keno:
I think @staticfloat said there were a few recipes in Yggdrasil that wanted this behavior. Probably needs an update. @staticfloat which ones were those?

staticfloat:
Looks like just one: https://github.com/JuliaPackaging/Yggdrasil/commit/32677499555e6920faa0e59eca433c09572ce8e3

PR #15, opened 2020-07-28
mkdir target directory when copying for DirectorySource

mortenpi:
I was running into an issue when my sources were just two `DirectorySources`, something like this:

```julia
sources = [
    DirectorySource("./src"), # should become ${WORKSPACE}/srcdir
    DirectorySource("./mysubdir", target="mysubdir"), # should become ${WORKSPACE}/srcdir/mysubdir
]
```

`build_tarballs` would then fail when it's copying over the second one with an exception on line 308, because it tries to copy into a non-existing directory.

Don't really know if this is the correct solution, or maybe my `sources` setup is just wrong, but it seems to work for me (TM).

### DIFF
--- a/src/Prefix.jl
+++ b/src/Prefix.jl
@@ -296,13 +296,12 @@ function setup(source::SetupSource{FileSource}, target, verbose)
 end
 
 function setup(source::SetupSource{DirectorySource}, targetdir, verbose)
+    mkpath(targetdir)
     # Need to strip the trailing separator also here
     srcpath = strip_backslash(source.path)
     if verbose
         @info "Copying content of $(basename(srcpath)) in $(basename(targetdir))..."
     end
-    # Create if missing. Will throw an IOError if targetdir is there, but not a directory.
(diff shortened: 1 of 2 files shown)


src/Prefix.jl
```diff
@@ -301,6 +301,8 @@ function setup(source::SetupSource{DirectorySource}, targetdir, verbose)
     if verbose
         @info "Copying content of $(basename(srcpath)) in $(basename(targetdir))..."
     end
+    # Create if missing. Will throw an IOError if targetdir is there, but not a directory.
+    isdir(targetdir) || mkdir(targetdir)
```
giordano, line 305: This should be equivalent to
```suggestion
    mkpath(targetdir)
```

mortenpi:
> Good catch, would you mind testing this feature as well?

Done :slightly_smiling_face: 

I decided to add it in as a separate test, so that it wouldn't change what the original one was testing.

giordano:
Thanks!!

PR #17, opened 2020-08-02
Add `GNU_LIBC_VERSION` to the mappings of the build environment

giordano:
I've looked into automatically determining the version of glibc a few times, but never found a general solution (only something that would work for `x86_64-linux-gnu`).  However some build systems insist on getting the version of glibc with `getconf GNU_LIBC_VERSION`, which however doesn't work, see for example https://github.com/SCIP-Interfaces/SCIP.jl/issues/165.  My proposal is to have an environment variable that at least gives the same result.  I guess someone might suggest to have a `getconf` script in the rootfs which simply does `echo ${@}`

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -726,6 +726,22 @@ function platform_envs(platform::Platform, src_name::AbstractString; host_platfo
         end
     end
 
+    # There is no easy way to automatically determine the version of glibc for
+    # all platforms, but some build systems want to know it.  Let's emulate with
+    # the `GNU_LIBC_VERSION` environment variable what `getconf
+    # GNU_LIBC_VERSION` would return, if it worked.
+    if libc(platform) === :glibc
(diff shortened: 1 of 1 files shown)


giordano:
For the record
```
echo '#include <errno.h>' | gcc -xc - -E -dM | grep -E '^#define __GLIBC(|_MINOR)__ '
```
should be a good way to programmatically get the version of Glibc, without having to compile and run a program (or execute libc.so).  For example
```console
% julia --compile=min -e 'using BinaryBuilderBase; BinaryBuilderBase.runshell(Platform("aarch64", "linux"))'
sandbox:${WORKSPACE} # echo '#include <errno.h>' | gcc -xc - -E -dM | grep -E '^#define __GLIBC(|_MINOR)__ '
#define __GLIBC__ 2
#define __GLIBC_MINOR__ 19
```

PR #32, opened 2020-09-12
Add keyword to `expand_cxxstring_abis` to skip FreeBSD and MacOS

giordano:
I'm not entirely sure this is a desirable feature because I think it'd make stuff like [`contrib/refresh_bb_tarballs.sh`](https://github.com/JuliaLang/julia/blob/eeffe578534289e4e99f72db263564173188b052/contrib/refresh_bb_tarballs.sh) more complicated.  However, in principle I believe this is a good thing to do, since when building for FreeBSD and macOS we usually don't link to libstdc++, and auditor would yell at us if there is something wrong.  I'm also happy to default `skip_freebsd_macos` to `false`

### DIFF
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -641,18 +641,21 @@ function expand_gfortran_versions(ps::Vector{<:Platform})
 end
 
 """
-    expand_cxxstring_abis(p::Platform)
+    expand_cxxstring_abis(p::Platform; skip_freebsd_macos::Bool=true)
 
 Given a `Platform`, returns an array of `Platforms` with a spread of identical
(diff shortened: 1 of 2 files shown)


giordano:
Ok, I'll fix the tests and add more if this PR is deemed to be acceptable, won't spend time on it otherwise :sweat_smile:

staticfloat:
I think this is a good change, actually.  Don't worry about complicating `refresh_bb_tarballs.sh` that is going away very soon.

giordano:
Tests adjusted, and fixed an actual bug in the original (untested) implementation

PR #36, opened 2020-09-21
Unshackle build-time dependency resolution from Pkg.Resolve

staticfloat:
This PR adds two things; first, usage of a new `Pkg` API to decouple
version resolution from constraints upon the currently-running version of
Julia, and the ability to set a version to be used at build-time that is
independent of the compat entries we will put into the generated JLL's
`Project.toml`.

### DIFF
--- a/src/Prefix.jl
+++ b/src/Prefix.jl
@@ -388,11 +388,17 @@ function setup_dependencies(prefix::Prefix, dependencies::Vector{PkgSpec}, platf
     end
     dependencies = filter_redundant_version.(dependencies)
 
+    # Get julia version specificity, if it exists, from the `Platform` object
+    julia_version = nothing
+    if haskey(platform, "julia_version")
+        julia_version = VersionNumber(platform["julia_version"])
+    end
+
     # We're going to create a project and install all dependent packages within
     # it, then create symlinks from those installed products to our build prefix
 
     # Update registry first, in case the jll packages we're looking for have just been registered/updated
-    ctx = Pkg.Types.Context()
+    ctx = Pkg.Types.Context(;julia_version = julia_version)
     outs = verbose ? stdout : devnull
     update_registry(ctx, outs)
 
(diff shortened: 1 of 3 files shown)


src/Prefix.jl
```diff
@@ -388,22 +388,27 @@ function setup_dependencies(prefix::Prefix, dependencies::Vector{PkgSpec}, platf
     end
     dependencies = filter_redundant_version.(dependencies)
 
+    # Get julia version specificity, if it exists, from the `Platform` object
+    julia_version = nothing
+    if haskey(platform, "julia_version")
+        julia_version = p["julia_version"]
+    end
```
giordano, line 395:
```suggestion
    julia_version = get(platform, "julia_version", nothing)
```
?  Is the value of `julia_version` a string or a `VersionNumber`?

staticfloat: Oh, that's a good point, it needs to be a `VersionNumber` for `Pkg`.

giordano:
Any good test for this feature?

staticfloat:
Hmmm, yeah, we could look at the versions that get pulled down when installing `GMP` versions across Julia versions.

giordano:
This caused https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/47

src/Prefix.jl
```diff
@@ -388,22 +388,27 @@ function setup_dependencies(prefix::Prefix, dependencies::Vector{PkgSpec}, platf
...
+    ctx = Pkg.Types.Context(;julia_version = julia_version)
     outs = verbose ? stdout : devnull
     update_registry(ctx, outs)
 
     mkpath(joinpath(prefix, "artifacts"))
     deps_project = joinpath(prefix, ".project")
     Pkg.activate(deps_project) do
         # Add all dependencies
-        Pkg.add(dependencies; platform=platform, io=outs)
+        Pkg.add(ctx, dependencies; platform=platform, io=outs)
```
giordano, line 409: This is adding packages to the global environment

PR #39, opened 2020-09-25
More correctly unmount shards in `DockerRunner`

staticfloat:


giordano:
Any quick way to test this?

staticfloat:
Mmmm, not unless we want to build and run the docker image during the tests here as well.

PR #43, opened 2020-10-01
Add HostBuildDependency

### LINKED ISSUE #37: Support HostBuildDependecies
For https://github.com/JuliaPackaging/Yggdrasil/pull/1694 it would be useful to be able to download the host version of `LLVM_full_jll` so that I can use `llvm-config` as the `LLVM_CONFIG_HOST` and get the right build flags.

giordano:
Main changes:

* we now have two "prefixes", one for the target and one for the host.  To
  implement this,
  * `setup_workspace` takes in input the platforms of the target and the host,
    and it creates the symlink
    `/workspace/destdir` -> `/workspace/TARGET_TRIPLET/destdir`
  * the user will call `setup_dependencies` twice, once for the host and once
    for the target
* a new helper function `destdir(prefix, platform)` gives consistently the path
  to the destdir for the given platform within the given prefix
* added relevant environment variables `host_*dir` in the runners, and add
  `host_bindir` to the `PATH`
* added `is_host_dependency`, `is_target_dependency`, `is_build_dependency`,
  `is_runtime_dependency` functions which are _very_ useful to filter
  dependencies in different situations: as we add more and more of these
  `AbstractDependency` types, there are multiple types matching each trait and
  it's hard to consistently list all of them in all places

CC: @vchuravy.  Fix #37

### DIFF
--- a/src/Dependencies.jl
+++ b/src/Dependencies.jl
@@ -1,6 +1,8 @@
 using UUIDs
 
-export Dependency, BuildDependency
+export Dependency, BuildDependency, HostBuildDependency,
+    is_host_dependency, is_target_dependency, is_build_dependency, is_runtime_dependency
+
 
 # Pkg.PackageSpec return different types in different Julia versions so...
 const PkgSpec = typeof(Pkg.PackageSpec(name="dummy"))
@@ -15,8 +17,39 @@ Concrete subtypes of `AbstractDependency` are
   and to load the generated JLL package.
 * [`BuildDependency`](@ref): a JLL package that is necessary only to build the
   package.  This will not be a dependency of the generated JLL package.
+* [`HostBuildDependency`](@ref): similar to `BuildDependency`, but it will
+  install the artifact for the host platform, instead of that for the target
+  platform.
 """
 abstract type AbstractDependency end
+
+"""
+    is_host_dependency(dep::AbstractDependency) -> Bool
+
+Return whether `dep` is a dependency of the host platform or not.
+"""
+is_host_dependency
+
+"""
+    is_target_dependency(dep::AbstractDependency) -> Bool
+
+Return whether `dep` is a dependency of the target platform or not.
+"""
+is_target_dependency(dep::AbstractDependency) = !is_host_dependency(dep)
+
+"""
+    is_build_dependency(dep::AbstractDependency) -> Bool
+
+Return whether `dep` is a build-time dependency or not.
+"""
+is_build_dependency
+
+"""
+    is_runtime_dependency(dep::AbstractDependency) -> Bool
+
+Return whether `dep` is a runtime dependency or not.
+"""
+is_runtime_dependency
 
 """
     Dependency(dep::Union{PackageSpec,String})
@@ -31,6 +64,9 @@ struct Dependency <: AbstractDependency
     Dependency(pkg::PkgSpec, build_version = nothing) = new(pkg, build_version)
 end
 Dependency(dep::AbstractString, build_version = nothing) = Dependency(PackageSpec(; name = dep), build_version)
+is_host_dependency(::Dependency) = false
+is_build_dependency(::Dependency) = true
+is_runtime_dependency(::Dependency) = true
 
 """
     BuildDependency(dep::Union{PackageSpec,String})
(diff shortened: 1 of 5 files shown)


giordano:
The reason why this is stalling is because moving the prefix will break the pkgconfing files, which are heavily used through the ecosystem.  They are like this:
```
prefix=/workspace/destdir
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: Xfixes
Description: X Fixes  Library
Version: 5.0.3
Requires: xproto fixesproto >= 5.0
Requires.private: x11
Cflags: -I${includedir}
Libs: -L${libdir} -lXfixes
```

giordano:
Reminder to self: need to add tests for `setup_dependencies`

giordano:
Companion PR for BinaryBuilder: https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/995

giordano:
This is now ready for review, together with its companion in BinaryBuilder!  I believe I fixed all remaining issues in `BinaryBuilder`.  The PR is still marked as draft because I want to add tests to https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/995 before merging this

src/Runner.jl
```diff
@@ -795,6 +808,12 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
         "libdir" => libdir,
         "includedir" => "$(prefix)/include",
 
+        # Set variables for the host prefix
+        "host_prefix" => host_prefix,
+        "host_bindir" => "$(host_prefix)/bin",
```
giordano, line 813: Should `host_bindir` be added to the `PATH`, maybe before `"$(prefix)/bin"`?

src/Dependencies.jl
```diff
@@ -43,6 +79,27 @@ struct BuildDependency <: AbstractDependency
...
 BuildDependency(dep::AbstractString) = BuildDependency(PackageSpec(; name = dep))
+is_host_dependency(::BuildDependency) = false
+is_build_dependency(::BuildDependency) = true
+is_runtime_dependency(::BuildDependency) = false
+
+"""
+    HostBuildDependency(dep::Union{PackageSpec,String})
+
+Define a binary dependency that is necessary only to build the package.
+Differently from the [`BuildDependency`](@ref), the artifact for the host
```
staticfloat, line 90:
```suggestion
Different from the [`BuildDependency`](@ref), the artifact for the host
```

src/Dependencies.jl
```diff
@@ -137,16 +194,18 @@ end
...
-        version = version == v"0" ? nothing : version
-        return BuildDependency(PackageSpec(; name = d["name"], uuid = uuid, version = version))
+        T = if d["type"] == "dependency"
+            Dependency
+        elseif d["type"] == "builddependency"
+            BuildDependency
+        elseif d["type"] == "hostdependency"
+            HostBuildDependency
+        end
+        return T(PackageSpec(; name = d["name"], uuid = uuid, version = version))
```
staticfloat, line 208: I had a moment of deja vu, where I was like "wait, I remember seeing this code before", then I realized that I implemented it the same way in my branch.  😂

src/Prefix.jl
```diff
@@ -341,10 +346,12 @@ function setup_workspace(build_path::AbstractString, sources::Vector;
...
     # We now set up two directories, one as a source dir, one as a dest dir
     srcdir = joinpath(workspace, "srcdir")
-    destdir = joinpath(workspace, "destdir")
+    target_destdir = destdir(workspace, target_platform)
+    host_destdir = destdir(workspace, host_platform)
     metadir = joinpath(workspace, "metadir")
-    wrapperdir = joinpath(workspace, "compiler_wrappers")
-    mkdir.((srcdir, destdir, metadir))
+    mkpath.((srcdir, target_destdir, host_destdir, metadir))
+    # Create the symlink /workspace/destdir -> /workspace/TARGET_TRIPLET/destdir
```
staticfloat, line 353:
```suggestion
    # Create the symlink /workspace/destdir -> /workspace/TARGET_TRIPLET/destdir
    # Necessary for compatibility with recipes that hardcode `/workspace/destdir` in them,
    # as well as `.pc` files that contain absolute paths to `/workspace/destdir/...`
```

As a side-note, I just learned about the `${pcfiledir}` variable in `.pc` files; you can use that to set up properly-relocatable `.pc` files by setting things like:

```
prefix=${pcfiledir}/../..
libdir=${prefix}/lib
...
```

So we should probably write an Audit pass to do that; would solve a lot of our `.pc` problems.  Issue here: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/108

staticfloat (approved):

PR #48, opened 2020-10-04
Add dependencies using the build environment instead of the global one

### LINKED ISSUE #47: BB uses global environment instead of creating a temporary
I've run locally this builder for [`OpenAL`](https://github.com/JuliaPackaging/Yggdrasil/blob/c716e7a90ff91f6dbfe36c72dbd47e8c92f9c2ad/O/OpenAL/build_tarballs.jl), which depends only on `alsa_jll`, `Ogg_jll` and `CompilerSupportLibraries_jll`.  None of them has further dependencies, but I've got installed in the build environment also completely unrelated packages, like `HDF5_jll`, `FFTW_jll`, `MKL_jll`, all MPI flavours, `Zlib_jll`, etc.... I realised these packages are all in my global environment.  After the build I found that my global environment has `alsa_jll`, `Ogg_jll` and `CompilerSupportLibraries_jll` installed.

I've bisected the issue to #36.

giordano:
Create the context after activating the temporary environment, otherwise we'd be
using the global one.

Fix #47.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,9 +1,7 @@
 # This file is machine-generated - editing it directly is not advised
 
 [[ArgTools]]
-git-tree-sha1 = "bdf73eec6a88885256f282d48eafcad25d7de494"
 uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
-version = "1.1.1"
 
 [[Artifacts]]
 uuid = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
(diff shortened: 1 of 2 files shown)


vchuravy:
Should we not just use a throw away environment?

giordano:
That's what we were supposed to be doing, but activating this environment https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/26a3194d2cb3717a10c64073f60b87ad2abcd55d/src/Prefix.jl#L407 but then this `Pkg.add` https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/26a3194d2cb3717a10c64073f60b87ad2abcd55d/src/Prefix.jl#L409 adds packages to this context https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/26a3194d2cb3717a10c64073f60b87ad2abcd55d/src/Prefix.jl#L401 which is the global environment.  This PR makes the context point to the  throw-away environment, at least this is my goal :sweat_smile:

giordano:
The culprit is this innocent change: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/36#discussion_r499295107.  By adding the `ctx` argument the packages are installed to a different environment than the one just activated.

vchuravy (approved):

PR #51, opened 2020-10-07
Fix JSON serialization for Dependency with build_version

### LINKED ISSUE #45: `Dependency("package_jll", v"1.2.3")` results in inconsistent compat bounds
https://github.com/JuliaPackaging/Yggdrasil/pull/1718 Used the new syntax `Dependency("GMP_jll", v"6.1.2")`, which however results in inconsistent compat bounds between `Project.toml` and the registry:

* `Project.toml` has a strict dependency on `v6.1.2:` https://github.com/JuliaBinaryWrappers/Singular_jll.jl/blob/85929b66afd6fdd6b877ec60fe8de9fe12661b27/Project.toml#L18
* the registry has a lower bound compat on `v6.1.2`: https://github.com/JuliaRegistries/General/pull/22388/files#diff-190a13cffbbf470e5e0c7f62902925bcR2

The registry here is what we want and the only thing that actually matters, but it's probably better to make these two bounds consistent

fingolfin:
Fixes #45

### DIFF
--- a/test/dependencies.jl
+++ b/test/dependencies.jl
@@ -22,6 +22,13 @@ end
     @test getname(dep) == name
     @test getname(PackageSpec(; name = name)) == name
     @test getpkg(dep) == PackageSpec(; name = name)
+
+    build_version = v"1.2.3"
+    dep_buildver = Dependency(PackageSpec(; name = name), build_version)
+    @test Dependency(name, build_version) == dep_buildver
+    @test getname(dep_buildver) == name
+    @test getpkg(dep_buildver) == PackageSpec(; name = name, version = build_version)
+
     build_name = "Foo_headers_jll"
     build_dep = BuildDependency(PackageSpec(; name = build_name))
     @test BuildDependency(build_name) == build_dep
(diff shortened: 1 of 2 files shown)


giordano:
Can you add tests please? :slightly_smiling_face:

fingolfin:
@giordano tests added

test/dependencies.jl
```diff
@@ -33,9 +40,16 @@ end
         jdep = JSON.lower(dep)
         @test jdep == Dict("type" => "dependency", "name" => name, "uuid" => nothing, "version-major" => 0x0, "version-minor" => 0x0, "version-patch" => 0x0)
         @test dependencify(jdep) == dep
+
+        jdep_buildver = JSON.lower(dep_buildver)
+        @test jdep_buildver == Dict("type" => "dependency", "name" => name, "uuid" => nothing, "version-major" => 0x0, "version-minor" => 0x0, "version-patch" => 0x0)
+        # the build_version is currently not serialized, so the following test fails
+        @test_broken dependencify(jdep_buildver) == dep_buildver
```
giordano, line 47: Should we make this work instead?

fingolfin: I am not sure (which is why I left it as-is). I can't think of a usecase for it right now, though, and my feeling was that code I don't write also is code that won't contain bugs ;-) But perhaps @staticfloat can think of a reasons why we might want to serialize the build version?

giordano: Ok, I don't think this is super-important now, I'm going to merge

staticfloat: Hmmm, it could be important in case we ever want to e.g. know precisely which version of a JLL was pushed up.  We should probably just change whatever preprocessing we're doing to `VersionNumbers` before serializing them to include the build number.

PR #52, opened 2020-10-13
Add `experimental` kwarg to `supported_platforms()`

staticfloat:
This will make it easier to have base julia deps always build for the
widest possible set of platforms, while letting the rest of Yggdrasil
continue to live in blissful ignorance

### DIFF
--- a/test/compat.jl
+++ b/test/compat.jl
@@ -44,7 +44,7 @@ using BinaryBuilderBase: download_verify, list_tarball_files
         @test_throws ErrorException download_verify("https://github.com/not_a_file", "0"^64, joinpath(dir, "blah"))
 
         # Test that a bad hash logs a message and fails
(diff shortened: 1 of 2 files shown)


test/compat.jl
```diff
@@ -44,7 +44,7 @@ using BinaryBuilderBase: download_verify, list_tarball_files
         @test_throws ErrorException download_verify("https://github.com/not_a_file", "0"^64, joinpath(dir, "blah"))
 
         # Test that a bad hash logs a message and fails
-        @test_logs (:error, r"Hash Mismatch") @test_throws ErrorException begin
+        @test_logs (:error, r"Hash Mismatch") match_mode=:any @test_throws ErrorException begin
```
giordano, line 47: This isn't necessary with #53

staticfloat: It's fine to keep it, IMO, since we don't really care if other logs are generated here.

PR #54, opened 2020-10-16
Use `curl` temporarily until `Downloads` fixes make their way into `1.6`

staticfloat:

PR #56, opened 2020-10-19
Handle existing files while downloading ourselves.

maleadt:
Seems more robust than relying on Downloads.jl not to redownload, and helps in the case that we can't even `HEAD` the URL (as, e.g., with NVIDIA's secure server; I just SCP the archives to the download server manually for those JLLs).

### DIFF
--- a/src/ArchiveUtils.jl
+++ b/src/ArchiveUtils.jl
@@ -127,6 +127,11 @@ function verify(path::AbstractString, hash::AbstractString; hash_path::AbstractS
 end
 
 function download_verify(url, hash, path)
+    if isfile(path)
+        verify(path, hash) && return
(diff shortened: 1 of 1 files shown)


maleadt:
CI failures unrelated, and happen locally on master too (i.e, without these changes).

giordano:
Yeah https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/54#issuecomment-711940197

giordano:
I think this is done by #59.  Here you are deleting the file, but I don't think it's really necessary and we can just happily overwrite it?

maleadt:
https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/59...

giordano:
It turns out `curl` doesn't overwrite existing files, so we need to explicitly delete them

PR #57, opened 2020-10-19
Mark currently broken tests as such

giordano:
All JLL packages formally depends on `Pkg`, including the new ones that don't
actually load it.  However, in Julia nightly `Pkg` has an indirect dependency on
`LibCURL_jll` and `MozillaCACerts_jll`, which means that `installed_jlls` in
`setup_dependencies` _always_ contains these two packages, see
https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/c4c7c838ad8cc2b134d0a46551497ef92f9a9e68/src/Prefix.jl#L411-L414

This changeset is to temporarily ignore the failing tests, for the sake of
keeping the test suite working, until this is properly resolved.

CC: @staticfloat


staticfloat (approved):

giordano:
Closing in favour of #58

PR #58, opened 2020-10-19
Filter out JLL packages that are in manifest only as stdlibs

giordano:


src/Prefix.jl
```diff
@@ -413,6 +432,16 @@ function setup_dependencies(prefix::Prefix, dependencies::Vector{PkgSpec}, platf
...
 
+        # Some JLLs are also standard libraries that are always present in the
+        # manifest.  Filter them out if they're present in the manifest but
+        # aren't direct dependencies or dependencies of other JLLS.
+        manifest = TOML.parsefile(ctx.env.manifest_file)
+        for jll in filter(s -> endswith(s, "_jll"), collect(values(Pkg.Types.stdlibs())))
+            if jll ∉ getfield.(dependencies, :name) && !is_dependency_of_jlls(jll, manifest)
+                filter!(p -> p.name != jll, installed_jlls)
+            end
+        end
```
staticfloat, line 443: So this says "for each JLL that is an stdlib, drop it unless it is either a direct dependency, or some other JLL depends on it".  But what about e.g. `LibCURL_jll` which depends on `nghttp2_jll`?  `is_dependency_of_jlls("nghttp2_jll", manifest)` will be `true`, I think.

I think what we need to do is walk the chain of deps from the root, and ensure that `is_dependency_of_jlls()` only returns `true` for something that has a chain of dependencies from the root to the bottom, all JLL packages.  E.g. `<root> -> LibCURL_jll -> nghttp2_jll` is allowable, while `<root> -> Foo_jll -> Pkg -> LibCURL_jll -> nghttp2_jll` is not allowable, because `Pkg` breaks the chain of JLLs.

I think we can eagerly compute this by doing a breadth-first search of the manifest, recording all JLL names that are unbroken chains, then we can filter out any JLLs that aren't in that set.

Manifest.toml
```diff
@@ -19,18 +19,19 @@ version = "0.7.0"
...
 uuid = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
 
 [[InteractiveUtils]]
 deps = ["Markdown"]
 uuid = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 
+[[JLLWrappers]]
+git-tree-sha1 = "7cec881362e5b4e367ff0279dd99a06526d51a55"
+uuid = "692b3bcd-3c85-4b1f-b108-f13ce0eb3210"
+version = "1.1.2"
```
staticfloat, line 33: You just can't stop pumping your numbers up, can you.  :P

staticfloat (changes requested):

src/Prefix.jl
giordano: Right, this works right now because the two JLLs don't have this issue right now :persevere:

PR #59, opened 2020-10-20
Do not download again sources that are already cached

giordano:
Fix https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/928

### DIFF
--- a/src/ArchiveUtils.jl
+++ b/src/ArchiveUtils.jl
@@ -127,6 +127,11 @@ function verify(path::AbstractString, hash::AbstractString; hash_path::AbstractS
 end
 
 function download_verify(url, hash, path)
-    Downloads.download(url, path)
-    verify(path, hash) || error("Verification failed")
+    if isfile(path) && verify(path, hash)
+        @info "Cached file found in $(path)"
+    else
+        @info "Downloading $(url) to $(path)..."
+        Downloads.download(url, path)
+        verify(path, hash) || error("Verification failed")
+    end
 end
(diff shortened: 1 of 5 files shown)


src/Sources.jl
```diff
@@ -133,11 +133,15 @@ function download_source(source::T; verbose::Bool = false, downloads_dir = stora
         src_path = abspath(source.url)
 
         # And if this is a locally-sourced tarball, just verify
-        verify(src_path, source.hash)
+        verify(src_path, source.hash) || error("Verification failed")
```
giordano, line 136: To be honest, I have no idea why we have this branch.  I'm leaving it as it is, but I don't know when we have sources that are local files

staticfloat:
Ahhhhhh, of course.  The old `download_verify()` would not download if the file already exists.  🤦‍♂️

src/ArchiveUtils.jl
```diff
@@ -127,6 +127,11 @@ function verify(path::AbstractString, hash::AbstractString; hash_path::AbstractS
 end
 
 function download_verify(url, hash, path)
-    Downloads.download(url, path)
-    verify(path, hash) || error("Verification failed")
+    if isfile(path) && verify(path, hash)
+        @info "Cached file found in $(path)"
+    else
+        @info "Downloading $(url) to $(path)..."
```
giordano, line 133: I hope this logging messages are fine.  I know this is a bit verbose, but they're very useful for debugging and making sure we're doing what we're supposed to.

src/Sources.jl
giordano: Unsurprisingly, these are 2 of the only 3 lines missing in code coverage in this file :smile:

PR #65, opened 2020-10-26
Switch platform arch i686 to use -march=pentium4

fingolfin:
Before, we used -march=i686, but that does not include MMX, SSE, SSE2 support,
which Julia requires. By switching to pentium4 as base level, better optimized
code can be generated, and also linking against libjulia then work for
32bit binaries without extra compiler flags.

Alternative to https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/960, see also the discussion there.

At least I hope this is a viable alternative -- I didn't test it thoroughly to verify.

### DIFF
--- a/src/Platforms.jl
+++ b/src/Platforms.jl
@@ -87,7 +87,7 @@ const ARCHITECTURE_FLAGS = Dict(
     # Many compiler flags are the same across clang and gcc, store those in "common"
     "common" => Dict(
         "i686" => Dict(
-            "i686" => ["-march=i686", "-mtune=generic"],
+            "pentium4" => ["-march=pentium4", "-mtune=generic"],
             "prescott" => ["-march=prescott", "-mtune=prescott"],
         ),
         "x86_64" => Dict(
--- a/test/rootfs.jl
+++ b/test/rootfs.jl
@@ -65,7 +65,7 @@ using BinaryBuilderBase
         # Platform("armv6l", "linux"; libc="glibc", call_abi="eabihf", march="arm1176jzfs"),
         Platform("armv7l", "linux"; libc="glibc", call_abi="eabihf", march="armv7l"),
         Platform("armv7l", "linux"; libc="glibc", call_abi="eabihf", march="neonvfpv4"),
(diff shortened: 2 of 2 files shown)


fingolfin:
@staticfloat thoughts on this?

src/Platforms.jl
```diff
@@ -87,7 +87,7 @@ const ARCHITECTURE_FLAGS = Dict(
     # Many compiler flags are the same across clang and gcc, store those in "common"
     "common" => Dict(
         "i686" => Dict(
-            "i686" => ["-march=i686", "-mtune=generic"],
+            "i686" => ["-march=pentium4", "-mtune=generic"],
```
staticfloat, line 90:
```suggestion
            "pentium4" => ["-march=pentium4", "-mtune=generic"],
```

Let's just call it `pentium4` everywhere.

staticfloat (changes requested):
We should also change the auditor to call the base 32-bit intel architecture `pentium4` instead of `i686`.

src/Platforms.jl
fingolfin: Huh, OK... but wouldn't that requires lots of changes here and also elsewhere? And have a risk of breaking things in all kinds  of subtle and unsubtle ways? I really don't dare make such a change myself, as I don't understand the ramifications and the big picture, so I'd rather prefer if somebody else who feels more confident in this took care of it.

staticfloat: Sorry, I didn't mean changing the architecture family name; just this "base microarchitectural level".

In reality, now that [this Base PR](https://github.com/JuliaLang/julia/pull/38186) was merged, I believe the only other thing that needs to change is [the docstring here](https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/master/src/auditor/instruction_set.jl#L81)

staticfloat: And since that change is in `BinaryBuilder.jl` I think this PR can be merged after it goes green.

fingolfin:
I adjusted the tests, so CI now passes (except for Coveralls coverage, for some reason)

staticfloat:
Perfect!  Thanks Max!

PR #70, opened 2020-11-04
Update aarch64 compiler, add workaround for broken host GCC v11

staticfloat:
@giordano so the end of the investigation is that the version of GCC 11 that it was based off of is just broken in its C++ header searching, so I have decided instead to just jam a host GCC 9 together with the aarch64-darwin GCC 11 shard, which should be fine.

### DIFF
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -370,7 +370,7 @@ const available_gcc_builds = [
     GCCBuild(v"7.1.0", (libgfortran_version = v"4", libstdcxx_version = v"3.4.23", cxxstring_abi = "cxx11")),
     GCCBuild(v"8.1.0", (libgfortran_version = v"5", libstdcxx_version = v"3.4.25", cxxstring_abi = "cxx11")),
     GCCBuild(v"9.1.0", (libgfortran_version = v"5", libstdcxx_version = v"3.4.26", cxxstring_abi = "cxx11")),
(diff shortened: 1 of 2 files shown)


src/Rootfs.jl
```diff
@@ -370,7 +370,7 @@ const available_gcc_builds = [
     GCCBuild(v"7.1.0", (libgfortran_version = v"4", libstdcxx_version = v"3.4.23", cxxstring_abi = "cxx11")),
     GCCBuild(v"8.1.0", (libgfortran_version = v"5", libstdcxx_version = v"3.4.25", cxxstring_abi = "cxx11")),
     GCCBuild(v"9.1.0", (libgfortran_version = v"5", libstdcxx_version = v"3.4.26", cxxstring_abi = "cxx11")),
-    GCCBuild(v"11-iains", (libgfortran_version = v"5", libstdcxx_version = v"3.4.26", cxxstring_abi = "cxx11")),
+    GCCBuild(v"11.0.0-iains", (libgfortran_version = v"5", libstdcxx_version = v"3.4.26", cxxstring_abi = "cxx11")),
```
giordano, line 373: Is the libstdc++ version accurate?  According to https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html GCC 10 has 3.4.28

staticfloat: good catch

PR #71, opened 2020-11-06
Fix bootstrapping, rebuild PlatformSupport and GCC 11

staticfloat:
Bootstrapping had bitrotted a little bit, fix that and also rebuild PS
shards for aarch64-darwin so that CMake works.


staticfloat:
With these changes and new artifacts, I can actually build LLVM_full for `aarch64-darwin`! :D

giordano (approved):

PR #72, opened 2020-11-12
Move CI to GitHub Actions

giordano:
TODO:

* [x] figure out how docker works in GitHub Actions
* [x] send coverage to coveralls/codecov


giordano:
These lines https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/3f67e2fcff0d0c0c93366ae8fa8678b018be2d31/src/DockerRunner.jl#L211-L213 implicitly assume that if the user doesn't explicitly pass the I/O streams they are `Base.TTY`s, but this is false in GitHub Actions, as the I/O streams are `Base.PipeEndpoin`s, causing the Docker tests to fail.  Not sure what's the best solution.  Maybe if the stream is nothing we should check `Base.stdxxx`?

giordano:
It seems to work now :tada:  Now on to code coverage

staticfloat (approved):

PR #73, opened 2020-11-13
Pass `-mmacosx-version-min` on `clang` as well as `gcc`

staticfloat:
I originally wanted to add the ability for the `os_version` tag in the
target to change things automatically, but we need to slightly update
the compiler shards first (e.g. remove the `14` from the paths in the
compiler shard and whatnot).

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -241,6 +241,31 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
         return flags
     end
 
+    function min_macos_version_flag(p::AbstractPlatform)
+        # If no `os_version` is specified in `p`, default to the oldest we support in the Julia world,
+        # which is `10.8`, but if it is actually specified, then set that corresponding value.
+        #min_macos_version = something(os_version(p), v"14.0.0")
+
+        # Eventually, we'll take this in `os_version(p)`, but not just yet.  We need to fix the paths
+        # to the compiler shards first, since right now they have `14` at the end
+        min_macos_version = v"14.0.0"
+
+        kernel_to_macos = Dict(
+            12 => "10.8",
+            13 => "10.9",
+            14 => "10.10",
+            15 => "10.11",
+            16 => "10.12",
+            17 => "10.13",
+            18 => "10.14",
+            19 => "10.15",
+            20 => "11.0",
+        )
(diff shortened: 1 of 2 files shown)


src/Runner.jl
```diff
@@ -241,6 +241,22 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
...
+        # If no `os_version` is specified in `p`, default to the oldest we support in the Julia world,
+        # which is `10.8`, but if it is actually specified, then set that corresponding value.
+        #min_macos_version = something(os_version(p), v"14.0.0")
+
+        # Eventually, we'll take this in `os_version(p)`, but not just yet.  We need to fix the paths
+        # to the compiler shards first, since right now they have `14` at the end
+        min_macos_version = v"14.0.0"
+
+        # `os_version` stores the kernel version, we translate that here to 10.X version:
+        min_macos_version = min_macos_version.major - 6
```
giordano, line 254: I'm not sure I understand this translation

src/Runner.jl
```diff
@@ -241,6 +241,22 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
...
+
+        # Always ask for a minimum macOS version of 10.8, as is default for the whole Julia world
+        return "-mmacosx-version-min=10.$(min_macos_version)"
```
giordano, line 257: This is
```
"-mmacosx-version-min=10.8.0.0"
```
is that correct?

src/Runner.jl
staticfloat: Darwin kernel version `14` means macOS `10.8`.

src/Runner.jl
staticfloat: `min_macos_version` is not a VersionNumber here, it's an Int.

src/Runner.jl
giordano: Ok, but does the relation version - 6 always hold?

src/Runner.jl
giordano: Uh uh, you're right

src/Runner.jl
staticfloat: Ugh, no.  haha.

PR #80, opened 2020-11-24
[Runner] Create symlinks to work around issues in Rust toolchain

giordano:
Ref https://github.com/JuliaPackaging/Yggdrasil/pull/2148 and
https://github.com/JuliaPackaging/Yggdrasil/pull/2168.

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -662,6 +662,22 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
             write_wrapper(rustc, p, "$(t)-rustc")
             write_wrapper(rustup, p, "$(t)-rustup")
             write_wrapper(cargo, p, "$(t)-cargo")
+
+            # For FreeBSD and macOS we need to create an unversioned link for
+            # gcc because that's the linker our Rust toolchain expects:
+            # https://github.com/JuliaPackaging/Yggdrasil/blob/fff0583bc2d8f32e450c427684f295524f38535d/0_RootFS/Rust/build_tarballs.jl#L115-L126.
+            if Sys.isbsd(p) && os_version(p) !== nothing
+                tmp_p = deepcopy(p)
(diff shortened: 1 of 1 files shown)


staticfloat:
Is this still necessary with the fixes to Rust?

giordano:
I think https://github.com/JuliaPackaging/Yggdrasil/pull/2168 is valuable, but it only fixes the ARM platforms.  We'd need to hard-code in the Rust toolchain the versions of FreeBSD and macOS to fix the other issue, but that seems fragile.

This PR has a workaround also for the issue with the 32-bit ARM platforms, so there is no need to rush to build a new Rust toolchain with the fix.

src/Runner.jl
```diff
@@ -662,6 +663,21 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
             write_wrapper(rustc, p, "$(t)-rustc")
             write_wrapper(rustup, p, "$(t)-rustup")
             write_wrapper(cargo, p, "$(t)-cargo")
+
+            # For FreeBSD and macOS we need to create an unversioned link for
+            # gcc because that's the linker our Rust toolchain expects:
+            # https://github.com/JuliaPackaging/Yggdrasil/blob/fff0583bc2d8f32e450c427684f295524f38535d/0_RootFS/Rust/build_tarballs.jl#L115-L126.
+            if Sys.isbsd(p) && os_version(p) !== nothing
+                delete!(tags(p), "os_version")
```
staticfloat, line 671: I'd feel more comfortable if we created a new `p` object here instead of mutating the existing one; we may want to use `p` in the future, and it would be bad if we didn't have the `os_version` tag when we expected to have it.

staticfloat (changes requested):

src/Runner.jl
giordano: Fair enough!

staticfloat (approved):

PR #82, opened 2020-11-25
[Runner] Move target wrappers to `/opt/bin/$(triplet(platform))`

### LINKED ISSUE #81: Cannot build for the host in `x86_64-linux-musl-cxx03`
We already know that host and target may step on each other toes: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/30d056ef68f81dca9cb91ededcce6b68c6466b37/src/Runner.jl#L586-L590  In https://github.com/JuliaPackaging/Yggdrasil/pull/2157#issuecomment-732989069 we wanted to build an executable for the host, but then it failed to run because we were actually building for the target:
```
% julia -e 'using BinaryBuilderBase; BinaryBuilderBase.runshell(Platform("x86_64", "linux"; libc="musl", cxxstring_abi="cxx03");preferred_gcc_version=v"5")'
sandbox:${WORKSPACE} # echo 'int main() {return 0;}'|SUPER_VERBOSE=1 ${CC_FOR_BUILD} -x c -
ccache /opt/x86_64-linux-musl/bin/x86_64-linux-musl-gcc -D_GLIBCXX_USE_CXX11_ABI=0 -frandom-seed=0x5ccd0c16 -march=x86-64 -mtune=generic -x c -
```
Not sure there is much we can do within the current setup (maybe let us choose the C++ string ABI with an environment variable?), but opening this for future reference

giordano:
This allows us to better discriminate between the wrappers for the host and
those for the target, important when the two platforms are similar (e.g., they
differ only by the C++ string ABI).

With this PR I can build https://github.com/JuliaPackaging/Yggdrasil/pull/2157 for `x86_64-linux-musl-cxx03` without any change.  Fix #81.

### DIFF
--- a/src/BuildToolchains.jl
+++ b/src/BuildToolchains.jl
@@ -1,0 +1,212 @@
+abstract type AbstractBuildToolchain{C} end
+
+struct CMake{C} <: AbstractBuildToolchain{C} end
+struct Meson{C} <: AbstractBuildToolchain{C} end
+
+c_compiler(::AbstractBuildToolchain{:clang}) = "clang"
+cxx_compiler(::AbstractBuildToolchain{:clang}) = "clang++"
+c_compiler(::AbstractBuildToolchain{:gcc}) = "gcc"
+cxx_compiler(::AbstractBuildToolchain{:gcc}) = "g++"
+fortran_compiler(::AbstractBuildToolchain) = "gfortran"
+
+function cmake_arch(p::AbstractPlatform)
+    if arch(p) == "powerpc64le"
+        return "ppc64le"
+    else
+        return arch(p)
+    end
+end
+
+function cmake_os(p::AbstractPlatform)
+    if Sys.islinux(p)
+        return "Linux"
+    elseif Sys.isfreebsd(p)
+        return "FreeBSD"
+    elseif Sys.isapple(p)
+        return "Darwin"
+    elseif Sys.iswindows(p)
+        return "Windows"
+    end
+end
+
+function toolchain_file(bt::CMake, p::AbstractPlatform)
+    target = triplet(p)
+    aatarget = aatriplet(p)
+
+    if Sys.isapple(p)
+        darwin_ver = something(os_version(p), v"14.5.0")
+        maj_ver = darwin_ver.major
+        min_ver = darwin_ver.minor
+        return """
+        # CMake toolchain file for $(c_compiler(bt)) running on $(target)
+        set(CMAKE_SYSTEM_NAME $(cmake_os(p)))
+        set(CMAKE_SYSTEM_PROCESSOR $(cmake_arch(p)))
+        set(CMAKE_SYSTEM_VERSION $(maj_ver).$(min_ver))
+        set(DARWIN_MAJOR_VERSION $(maj_ver))
+        set(DARWIN_MINOR_VERSION $(min_ver))
+
+        # Enable rpath support
+        set(CMAKE_SHARED_LIBRARY_RUNTIME_C_FLAG "-Wl,-rpath,")
+
+        set(CMAKE_SYSROOT /opt/$(aatarget)/$(aatarget)/sys-root)
+        set(CMAKE_SYSTEM_FRAMEWORK_PATH
+            \${CMAKE_SYSROOT}/System/Library/Frameworks
+            \${CMAKE_SYSROOT}/System/Library/PrivateFrameworks
+        )
+        set(CMAKE_INSTALL_PREFIX \$ENV{prefix})
+
+        set(CMAKE_C_COMPILER   /opt/bin/$(target)/$(c_compiler(bt)))
+        set(CMAKE_CXX_COMPILER /opt/bin/$(target)/$(cxx_compiler(bt)))
+        set(CMAKE_Fortran_COMPILER /opt/bin/$(target)/$(fortran_compiler(bt)))
+
+        set(CMAKE_LINKER  /opt/bin/$(target)/$(aatarget)-ld)
+        set(CMAKE_OBJCOPY /opt/bin/$(target)/$(aatarget)-objcopy)
+
+        set(CMAKE_AR     /opt/bin/$(target)/$(aatarget)-ar)
+        set(CMAKE_NM     /opt/bin/$(target)/$(aatarget)-nm)
+        set(CMAKE_RANLIB /opt/bin/$(target)/$(aatarget)-ranlib)
+
+        if( \$ENV{CC} MATCHES ccache )
+            set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
+        endif()
+        """
+    else
+        return """
+        # CMake toolchain file for $(c_compiler(bt)) running on $(target)
+        set(CMAKE_SYSTEM_NAME $(cmake_os(p)))
+        set(CMAKE_SYSTEM_PROCESSOR $(cmake_arch(p)))
+
+        set(CMAKE_SYSROOT /opt/$(aatarget)/$(aatarget)/sys-root/)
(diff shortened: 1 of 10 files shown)


src/Runner.jl
```diff
@@ -925,17 +927,21 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
     # so we set all the environment variables that we've seen them called
     # and hope for the best.
     for host_map in (tool -> "HOST$(tool)", tool -> "$(tool)_FOR_BUILD", tool -> "BUILD_$(tool)", tool -> "$(tool)_BUILD")
+        # Use full path to avoid collisions when the target is similar to the
+        # host (e.g., `x86_64-linux-musl-cxx03` and `x86_64-linux-musl-cxx11`)
+        host_bin_dir = "/opt/bin/$(aatriplet(host_platform))"
```
staticfloat, line 932: I think you don't want to use `aatriplet()` but instead just `triplet()`.  Since `aatriplet()` should strip out the abi parts.  (`aa` means `"abi-agnostic"`)

giordano: These are just environment variables, those paths come from the RootFS

giordano: It turned out these paths didn't come from the RootFS, but the fact that they were implicitly ABI agnostic was because currently the host platform is always `x86_64-linux-musl`, without specification of the ABI.

src/Runner.jl
```diff
@@ -846,7 +828,9 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
     merge!(mapping, Dict(
         "PATH" => join((
             # First things first, our compiler wrappers trump all
-            "/opt/bin",
+            "/opt/bin/$(triplet(platform))",
+            # Also wrappers for the host
+            "/opt/bin/$(triplet(host_platform))",
```
giordano, line 833: I think you wanted to have also host wrappers in the PATH, right?  but maybe after all target wrappers?

test/rootfs.jl
```diff
@@ -145,50 +145,44 @@ end
...
-        end
-        platform      = Platform("x86_64", "linux"; libc="musl", cxxstring_abi="cxx03")
-        host_platform = Platform("x86_64", "linux"; libc="musl", cxxstring_abi="cxx11")
-        mktempdir() do bin_path
-            @test_throws ErrorException generate_compiler_wrappers!(platform; bin_path = bin_path, host_platform = host_platform)
-        end
-        platform      = Platform("x86_64", "linux"; libc="musl")
-        host_platform = Platform("x86_64", "linux"; libc="musl", cxxstring_abi="cxx03")
-        mktempdir() do bin_path
-            @test_throws ErrorException generate_compiler_wrappers!(platform; bin_path = bin_path, host_platform = host_platform)
```
giordano, line 191: These tests were checking the error at https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/30d056ef68f81dca9cb91ededcce6b68c6466b37/src/Runner.jl#L92 which is now removed

src/Runner.jl
```diff
@@ -69,31 +74,6 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
...
-        host_cxxabi   = cxxstring_abi(host_platform)
-        if target_cxxabi !== host_cxxabi
-            if host_cxxabi !== nothing
-                # This is a very unlikely situation as ideally the host
-                # shouldn't have particular preferences for the ABI, thus in
-                # practice we should never reach this.
-                error("Incompatible C++ string ABIs between the host and target platforms")
-            end
-        end
-    end
```
giordano, line 95: We're now correctly separating the wrappers between the host and the target, so we don't really need this guard anymore

giordano:
This breaks CMake (and probably also meson) :disappointed:

staticfloat:
Presumably we need to generate new `PlatformSupport` shards with the updated paths?

giordano:
Yes, and this should be done together with this

giordano:
Apparently we can't reference environment variables in the meson cross files: https://github.com/mesonbuild/meson/issues/309...

giordano:
Ok, I've added a file which generates the toolchains files for CMake and Meson, much like what we do now for the compiler wrappers.  I've done a couple of simple tests locally, they seem to work

src/BuildToolchains.jl
```diff
@@ -0,0 +1,203 @@
...
+function toolchain_file(bt::CMake, p::AbstractPlatform)
+    target = triplet(p)
+    aatarget = aatriplet(p)
+
+    if Sys.isapple(p)
+        return """
+        # CMake toolchain file for $(c_compiler(bt)) running on $(target)
+        set(CMAKE_SYSTEM_NAME $(cmake_os(p)))
+        set(CMAKE_SYSTEM_PROCESSOR $(cmake_arch(p)))
+        set(CMAKE_SYSTEM_VERSION 14.5)
```
giordano, line 40: I think this is the version of Darwin?  Maybe we can use `os_version(p)` instead of hard-coding 14.5?

src/BuildToolchains.jl
```diff
@@ -0,0 +1,203 @@
...
+
+        # Extract out DARWIN_MAJOR_VERSION and DARWIN_MINOR_VERSION
+        string(REGEX REPLACE "^([0-9]+)\\\\.([0-9]+).*\$" "\\\\1" DARWIN_MAJOR_VERSION "\${CMAKE_SYSTEM_VERSION}")
+        string(REGEX REPLACE "^([0-9]+)\\\\.([0-9]+).*\$" "\\\\2" DARWIN_MINOR_VERSION "\${CMAKE_SYSTEM_VERSION}")
```
giordano, line 44: And we can automatically set this from `os_version(p)` instead of playing with regexps

src/BuildToolchains.jl
staticfloat: Agreed.  Let's use `os_version(p)`.

src/BuildToolchains.jl
staticfloat: Yes, agreed.  Wonderful!

src/Rootfs.jl
```diff
@@ -497,8 +497,8 @@ function choose_shards(p::AbstractPlatform;
             preferred_llvm_version::VersionNumber = getversion(LLVM_builds[end]),
         )
 
-    # Our host platform is x86_64-linux-musl
-    host_platform = Platform("x86_64", "linux"; libc="musl")
+    # Our host platform is x86_64-linux-musl-cxx11
+    host_platform = default_host_platform
```
staticfloat, line 501: Can we just replace usage of `host_platform` with `default_host_platform`?

src/Runner.jl
```diff
@@ -1,6 +1,11 @@
 import Base: strip
 abstract type Runner; end
 
+# Host platform _must_ match the C++ string ABI of the binaries we get from the
+# repositories.  Note: when preferred_gcc_version=v"4" we can't really build for
+# that C++ string ABI :-(
```
staticfloat, line 6:
```suggestion
# Host platform _must_ match the C++ string ABI of the binaries we get from the
# Alpine repositories, so that when using `$HOSTCC` to compile host executables,
# they can link against e.g. `libboost` provided through `apk`.
# Note: when preferred_gcc_version=v"4" we can't really build for that C++ string
# ABI, so GCC 5 will be chosen instead.
```

staticfloat (approved):
This is fantastic work.  Only minor comments and predominantly agreeing with all of your suggestions.

src/Runner.jl
giordano: > so GCC 5 will be chosen instead

That's not true unfortunately :disappointed:

src/Rootfs.jl
giordano: Good point, we aren't really allowing to use something else

PR #86, opened 2020-12-01
Roll `Rust` build back to `v1.43`

staticfloat:
This is required because of a regression in the `Rust` build on
i686-w64-mingw32: https://github.com/rust-lang/rust/issues/79609


giordano:
It's all green also in BinaryBuilder, thanks!

PR #93, opened 2020-12-21
Properly (de)serialize VersionSpec to JSON

### LINKED ISSUE #89: Meta JSON serialization does not preserve VersionSpec
The JSON serialization via `--meta-json` that is used during the Yggdrasil build will convert any VersionSpec to just a fixed full VersionNumber `x.y.z`, this is done here https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/573b396e475c05353ce9d975bd726774d59b8e39/src/Dependencies.jl#L127-L149
This happened for example in https://github.com/JuliaPackaging/Yggdrasil/pull/2235 where
```
Dependency(PackageSpec(name="FLINT_jll", version=VersionSpec("200.690"))),
```
resulted in
- [registry](https://github.com/JuliaRegistries/General/blob/b169fc4a38b81d6e6eec7760600420bac2d28ffd/S/Singular_jll/Compat.toml#L6) `FLINT_jll = "200.690.0"`
- [project file](https://github.com/JuliaBinaryWrappers/Singular_jll.jl/blob/22ac1b95da33c159789235d5e27a2f33b1e833e1/Project.toml#L18) `FLINT_jll = "=200.690.0"`

cc @fingolfin

fingolfin:
This changes `Dependency` objects to always (?) store a full `VersionSpec`, not just a `VersionNumber`, and (de)serialize that properly.  Fixes #89 (well, I hope).

Before this, https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/981 should be merged.

CC @benlorenz @thofma

### DIFF
--- a/src/Dependencies.jl
+++ b/src/Dependencies.jl
@@ -113,7 +113,7 @@ end
 # Add JSON serialization of dependencies
 string_or_nothing(x) = isnothing(x) ? x : string(x)
 
-# helper to get the version of a dependency (and no the build_version)
+# helper to get the version of a dependency (and not the build_version)
 version(d::AbstractDependency) = getpkg(d).version
 version(d::Dependency) = d.pkg.version
 
--- a/src/Dependencies.jl
+++ b/src/Dependencies.jl
@@ -113,43 +113,37 @@ end
 # Add JSON serialization of dependencies
 string_or_nothing(x) = isnothing(x) ? x : string(x)
 
-major(v::VersionNumber) = v.major
-minor(v::VersionNumber) = v.minor
-patch(v::VersionNumber) = v.patch
-major(v::Pkg.Types.VersionBound) = v.t[1]
-minor(v::Pkg.Types.VersionBound) = v.t[2]
-patch(v::Pkg.Types.VersionBound) = v.t[3]
-__version(v::VersionNumber) = v
-__version(v::Pkg.Types.VersionSpec) = v.ranges[1].lower
-version(d::AbstractDependency) = __version(getpkg(d).version)
-version(d::Dependency) = __version(d.pkg.version)
+# helper to get the version of a dependency (and no the build_version)
+version(d::AbstractDependency) = getpkg(d).version
+version(d::Dependency) = d.pkg.version
 
 for (type, type_descr) in ((Dependency, "dependency"), (BuildDependency, "builddependency"))
     JSON.lower(d::type) = Dict("type" => type_descr,
                                "name" => d.pkg.name,
                                "uuid" => string_or_nothing(d.pkg.uuid),
-                               "version-major" => major(version(d)),
-                               "version-minor" => minor(version(d)),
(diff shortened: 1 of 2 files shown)


src/Dependencies.jl
```diff
@@ -113,43 +113,37 @@ end
...
-major(v::Pkg.Types.VersionBound) = v.t[1]
-minor(v::Pkg.Types.VersionBound) = v.t[2]
-patch(v::Pkg.Types.VersionBound) = v.t[3]
-__version(v::VersionNumber) = v
-__version(v::Pkg.Types.VersionSpec) = v.ranges[1].lower
-version(d::AbstractDependency) = __version(getpkg(d).version)
-version(d::Dependency) = __version(d.pkg.version)
+# helper to get the version of a dependency (and no the build_version)
+version(d::AbstractDependency) = getpkg(d).version
+version(d::Dependency) = d.pkg.version
```
giordano, line 118: What's the difference?  Isn't the first method sufficient?

fingolfin: No it is (and was) not enough, which I tried to explain in the comment I added above

fingolfin: Any suggestions on how to make this clearer are welcome :-)

src/Dependencies.jl
```diff
@@ -113,43 +113,37 @@ end
...
-minor(v::VersionNumber) = v.minor
-patch(v::VersionNumber) = v.patch
-major(v::Pkg.Types.VersionBound) = v.t[1]
-minor(v::Pkg.Types.VersionBound) = v.t[2]
-patch(v::Pkg.Types.VersionBound) = v.t[3]
-__version(v::VersionNumber) = v
-__version(v::Pkg.Types.VersionSpec) = v.ranges[1].lower
-version(d::AbstractDependency) = __version(getpkg(d).version)
-version(d::Dependency) = __version(d.pkg.version)
+# helper to get the version of a dependency (and no the build_version)
```
fingolfin, line 116:
```suggestion
# helper to get the version of a dependency (and not the build_version)
```

src/Dependencies.jl
fingolfin: Just to make this explicit: this trick is not by me, it was added by @staticfloat as part of the build_version support; I merely dropped the (now unnecessary) `__version` and added a comment in an attempt to clarify what's going on (well, or at least give a hint as to what's going on). It can probably be phrased better (and I just noticed a typo).

giordano: I was missing that `getpkg(d::Dependency)` isn't as simple as `d.pkg`.  At this point I find a bit questionable that method, since this is basically undoing it.

fingolfin:
@staticfloat @giordano it would be really good to get this or something equivalent merged and used on Yggdrasil; it is starting to seriously bite me in the a... now

staticfloat:
Thanks Max!  We can fix up the `getpkg()` stuff later.  :)

PR #104, opened 2021-01-27
Fix MacOS GCC shards

staticfloat:
Our MacOS shards are now providing `dsymutil` as the primary name, so
just use that instead.

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -560,7 +560,7 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
             return (io, p) -> nothing
         end
 
-        return wrapper(io, "/opt/$(aatriplet(p))/bin/llvm-dsymutil"; allow_ccache=false)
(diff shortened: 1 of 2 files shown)


src/Runner.jl
```diff
@@ -948,7 +948,7 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
         for (env_name, tool) in (
             "CC" => "$(host_bin_dir)/$(host_target)-gcc",
             "CXX" => "$(host_bin_dir)/$(host_target)-g++",
-            "DSYMUTIL" => "llvm-dsymutil",
+            "DSYMUTIL" => "dsymutil",
```
omus, line 951: I noticed that `llvm-dsymutil` didn't exist while investigating an issue with building an experimental GCC 11. I just thought this was my issue. Things make more sense now!

giordano:
If https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/997 is green, we're good to go

giordano (approved):

Issue #106, opened 2021-01-28
Packaging GCCBootstrap too slow using 7zip Gzip compression


omus:
I've been experimenting with building the `GCCBootstrap@11-IainS` shard for `aarch64-apple-darwin20` (Apple Silicon) and have found that most of the time is taken up compressing the artifact. In particular when running on AWS using an `c5ad.8xlarge` instance I found that I could build the shard in 23 minutes but compressing the `GCCBootstrap.v11.0.0-iains.aarch64-apple-darwin20.tar.gz` was taking around an 1 hour and 10 minutes (plus another round of compression for squashfs).

When BinaryBuilder shows the output:
```
[ Info: Tree hash of contents of GCCBootstrap.v11.0.0-iains.aarch64-apple-darwin20.tar.gz: 120bcbd3d356f2fa86d3919c58ee1dd67ab757f5
```
You can see in the background the process [`7z a -si -tgzip -mx9`](https://github.com/JuliaLang/Pkg.jl/blob/76402b8974aba76f4412e59bee2e4bc6be1b7860/src/PlatformEngines.jl#L387) running on a single thread.

We could potentially greatly benefit from using an alternate tool for performing compression, specifically one that can use multiple threads.

omus:
I've looked into other compression tools ([`pigz`](https://zlib.net/pigz/) and [`zstd`](https://github.com/facebook/zstd)) and formats to try and speed up the process. I used a the uncompressed contents of `GCCBootstrap.v11.0.0-iains.aarch64-apple-darwin20.tar.gz` which uncompressed is a size of 5316188080 bytes (5.1GB) and with 5629 files. I performed these tests on an AWS `c5ad.8xlarge` instance using NVME storage.

```
$ tar cvf - . | 7za a -si -tgzip -mx9 ../7z9-test.tar.gz  #  No threading, using gzip at maximum compression
real  69m48.743s
user  69m44.884s
sys   0m6.908s
size  1558218029 bytes (1.5GB)

$ tar cvf - . | 7za a -si -tgzip -mx9 -mmt32 ../7z9T-test.tar.gz  #  Threading enabled, but deflate doesn't support it, using gzip at maximum compression
TBD, I expect same as above

$ tar cvf - . | 7za a -si -t7z -mx9 -mmt32 ../7z9T-test.tar.7z  # Threading enabled, using 7z at maximum compression
real  2m48.593s
user  53m14.795s
sys   0m34.856s
[...]

PR #107, opened 2021-01-28
Use `pigz` to create archives

omus:
Fixes https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/106. Using `pigz` allows us to perform fast Gzip compression. In the future we should probably switch to the `7z` or `zstd` format.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -28,9 +28,9 @@ deps = ["Markdown"]
 uuid = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 
 [[JLLWrappers]]
-git-tree-sha1 = "7cec881362e5b4e367ff0279dd99a06526d51a55"
+git-tree-sha1 = "a431f5f2ca3f4feef3bd7a5e94b8b8d4f2f647a0"
 uuid = "692b3bcd-3c85-4b1f-b108-f13ce0eb3210"
-version = "1.1.2"
+version = "1.2.0"
 
 [[JSON]]
 deps = ["Dates", "Mmap", "Parsers", "Unicode"]
@@ -47,7 +47,7 @@ deps = ["Libdl"]
 uuid = "deac9b47-8bc7-5906-a0fe-35ac56dc84c0"
 
 [[LibGit2]]
-deps = ["Printf"]
+deps = ["Base64", "NetworkOptions", "Printf", "SHA"]
 uuid = "76f85450-5226-5b5a-8eaa-529ad045b433"
 
 [[Libdl]]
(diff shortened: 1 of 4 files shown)


Manifest.toml
```diff
@@ -47,7 +47,7 @@ deps = ["Libdl"]
 uuid = "deac9b47-8bc7-5906-a0fe-35ac56dc84c0"
 
 [[LibGit2]]
-deps = ["Printf"]
+deps = ["Base64", "NetworkOptions", "Printf", "SHA"]
```
giordano, line 50: I don't think the Julia we use here likes `NetworkOptions`

omus: The Manifest was modified using Julia 1.6 (specifically commit a178ac2cba) so this should be good.

omus:
As part of my testing I confirmed that `tar` could decompress the archives created by `pigz` and that the tree hash remained identical.

src/ArchiveUtils.jl
```diff
@@ -146,3 +148,37 @@ function download_verify(url, hash, path)
         verify(path, hash) || error("Verification failed")
     end
 end
+
+# Copy of `Pkg.Artifacts.archive_artifact` that supports a custom `package` function
+function _archive_artifact(hash::SHA1, tarball_path::String;
+                           honor_overrides::Bool=false,
+                           package::Function=_package_fast)
+
+    if !honor_overrides
```
staticfloat, line 157: You can drop `honor_overrides`

staticfloat (approved):

src/ArchiveUtils.jl
omus: Sure, I just kept the function as close to the original as possible

src/Prefix.jl
```diff
@@ -173,7 +173,7 @@ function package(prefix::Prefix,
         @info("Tree hash of contents of $(basename(out_path)): $(tree_hash)")
     end
 
-    tarball_hash = archive_artifact(tree_hash, out_path; honor_overrides=false)
+    tarball_hash = _archive_artifact(tree_hash, out_path; honor_overrides=false)
```
staticfloat, line 176: need to remove it here too

omus: From your [original comment](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/107#discussion_r566263696) I assumed we weren't using this keyword. What's the reason behind removing this? We'll probably want to switch back to the `Pkg.Artifacts` version at some point in the future

omus: I appreciate the explanation. Leaving the `honor_overrides` kwarg in place sounds like the ideal solution to me.

PR #109, opened 2021-01-28
Rebuild PlatformSupport 2021.1.28

omus:
I believe I did this correctly but as I had some issues building these I want to do some additional verification. 
Associated Yggdrasil release: https://github.com/JuliaPackaging/Yggdrasil/releases/tag/PlatformSupport-v2021.1.28

Introduces PlatformSupport artifacts for `aarch64-apple-darwin20` and `i686-linux-musleabihf`.

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -2278,6 +2278,28 @@ os = "linux"
     sha256 = "7bc24b76da255ea583622169264a9c02aa80bc927a0a23bcbd4ed814a99fa629"
     url = "https://github.com/JuliaPackaging/Yggdrasil/releases/download/PlatformSupport-v2021.1.28/PlatformSupport-i686-linux-musl.v2021.1.28.x86_64-linux-musl.unpacked.tar.gz"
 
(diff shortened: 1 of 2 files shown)


giordano:
Is [this](https://github.com/JuliaPackaging/Yggdrasil/pull/2436) the corresponding PR in Yggdrasil?  Or there is something else?

omus:
I'm missing `i686-linux-musleabihf`

omus:
> Is [this](https://github.com/JuliaPackaging/Yggdrasil/pull/2436) the corresponding PR in Yggdrasil? Or there is something else?

Yes, should I have built off of that commit instead?

giordano:
No, I just wanted to have a reference of what we're getting here, we can't see much with this kind of commit :sweat_smile:

omus:
Comparing this to the previous release of `v2020.11.6` I just noticed that `i686-linux-musleabihf` wasn't included in the release. Seems like I discovered this when accidentally using `expand_cxxstring_abis` when building the various PlatformSupport targets. I can remove it from the listing here but it seemed fine to leave it in.

omus:
Ready to go from my end

PR #112, opened 2021-02-02
Add LLVMBoostrap 11.0.1

omus:
Ref: https://github.com/JuliaPackaging/Yggdrasil/pull/2473

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -2044,6 +2044,28 @@ os = "linux"
     sha256 = "b8488f009c117eb7135b2ac139009836518587aacea772eba83271c794afa420"
     url = "https://github.com/JuliaPackaging/Yggdrasil/releases/download/Go-v1.13.0+0/Go.v1.13.0.x86_64-linux-musl.unpacked.tar.gz"
 
+[["LLVMBootstrap.v11.0.1.x86_64-linux-musl.squashfs"]]
+arch = "x86_64"
+git-tree-sha1 = "5ab848b00d11e579786ffcca46f2ab0173e30d48"
+lazy = true
+libc = "musl"
+os = "linux"
+
+    [["LLVMBootstrap.v11.0.1.x86_64-linux-musl.squashfs".download]]
+    sha256 = "b33850954f65d988278b2a81d561bd3fdaa2c8e7bb08fcce553df1fc27cc9f8e"
+    url = "https://github.com/JuliaPackaging/Yggdrasil/releases/download/LLVMBootstrap-v11.0.1/LLVMBootstrap.v11.0.1.x86_64-linux-musl.squashfs.tar.gz"
+
+[["LLVMBootstrap.v11.0.1.x86_64-linux-musl.unpacked"]]
+arch = "x86_64"
+git-tree-sha1 = "0dbbe23941529c027bea16a585cb971f26ac058c"
+lazy = true
+libc = "musl"
+os = "linux"
+
+    [["LLVMBootstrap.v11.0.1.x86_64-linux-musl.unpacked".download]]
+    sha256 = "fc9a3d15aca56d332cfa3caeba1a7bd6482cf1a82ef55ba53419a976c6a4e9e3"
+    url = "https://github.com/JuliaPackaging/Yggdrasil/releases/download/LLVMBootstrap-v11.0.1/LLVMBootstrap.v11.0.1.x86_64-linux-musl.unpacked.tar.gz"
+
 [["LLVMBootstrap.v6.0.1.x86_64-linux-musl.squashfs"]]
 arch = "x86_64"
 git-tree-sha1 = "474b55ee699607660a7375f4e38afc9a76583c41"
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -389,6 +389,7 @@ const available_llvm_builds = [
     LLVMBuild(v"7.1.0"),
     LLVMBuild(v"8.0.1"),
     LLVMBuild(v"9.0.1"),
+    LLVMBuild(v"11.0.1"),
 ]
 
 """


giordano:
Can you please also update the [list of available llvm versions](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/940ce9b5e0844721468cf72ecdd3f4ebd039ad61/src/Rootfs.jl#L376)?

giordano:
Test failures don't look good:
```
Compilation - macOS x86_64: Test Failed at /home/runner/work/BinaryBuilderBase.jl/BinaryBuilderBase.jl/test/runners.jl:104
  Expression: (split(String(read(iobuff)), "\n"))[2] == ""
   Evaluated: "error: unknown target CPU 'x86-64'" == ""
Stacktrace:
 [1] (::var"#199#214"{Platform})(dir::String)
   @ Main ~/work/BinaryBuilderBase.jl/BinaryBuilderBase.jl/test/runners.jl:104
 [2] mktempdir(fn::var"#199#214"{Platform}, parent::String; prefix::String)
   @ Base.Filesystem ./file.jl:713
 [3] mktempdir(fn::Function, parent::String)
   @ Base.Filesystem ./file.jl:711
 [4] macro expansion
   @ ~/work/BinaryBuilderBase.jl/BinaryBuilderBase.jl/test/runners.jl:99 [inlined]
 [5] macro expansion
   @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1221 [inlined]
 [6] macro expansion
   @ ~/work/BinaryBuilderBase.jl/BinaryBuilderBase.jl/test/runners.jl:98 [inlined]
 [7] macro expansion
   @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1146 [inlined]
 [8] top-level scope
   @ ~/work/BinaryBuilderBase.jl/BinaryBuilderBase.jl/test/runners.jl:32

[...]

Compilation - FreeBSD x86_64: Test Failed at /home/runner/work/BinaryBuilderBase.jl/BinaryBuilderBase.jl/test/runners.jl:104
  Expression: (split(String(read(iobuff)), "\n"))[2] == ""
   Evaluated: "error: unknown target CPU 'x86-64'" == ""
Stacktrace:
 [1] (::var"#199#214"{Platform})(dir::String)
   @ Main ~/work/BinaryBuilderBase.jl/BinaryBuilderBase.jl/test/runners.jl:104
 [2] mktempdir(fn::var"#199#214"{Platform}, parent::String; prefix::String)
   @ Base.Filesystem ./file.jl:713
 [3] mktempdir(fn::Function, parent::String)
   @ Base.Filesystem ./file.jl:711
 [4] macro expansion
   @ ~/work/BinaryBuilderBase.jl/BinaryBuilderBase.jl/test/runners.jl:99 [inlined]
 [5] macro expansion
   @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1221 [inlined]
 [6] macro expansion
   @ ~/work/BinaryBuilderBase.jl/BinaryBuilderBase.jl/test/runners.jl:98 [inlined]
 [7] macro expansion
   @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1146 [inlined]
 [8] top-level scope
   @ ~/work/BinaryBuilderBase.jl/BinaryBuilderBase.jl/test/runners.jl:32
```
and since this happens for macOS and FreeBSD which use LLVM by default, this is likely a real issue

omus:
I'm unable to reproduce locally when using `BINARYBUILDER_FULL_SHARD_TEST=true`. I'll restart the CI just to ensure the issue is reproducible here at least.

omus:
I deleted my local copies of the LLVMBootstrap 11 artifacts and noticed they weren't re-downloaded during these tests. I suspect this is an unrelated failure.

omus:
🎉  I've managed to reproduce the issue locally. My local branch and pushed branch were out of sync

giordano:
> My local branch and pushed branch were out of sync

That's an interesting situation :smile:

omus:
After some experimentation I found that LLVM 10.0.1 doesn't have this problem while LLVM 11.0.0 does. Performing a bisect I found that https://github.com/llvm/llvm-project/commit/3537939cda86f0b5b06233eb99ddc9eb22935008 is the first commit that produces the issue.

giordano:
Good detective job! :+1:

vchuravy:
> that LLVM 10.0.1 doesn't have this problem while LLVM 11.0.0 does. 

I assume you meant the inverse?

omus:
After reviewing the failing commit and further changes I discovered https://github.com/llvm/llvm-project/commit/9ecda9aa804dcb49e30e46161b29976494b0a1f9 ([llvm-commits thread](https://lists.llvm.org/pipermail/llvm-commits/Week-of-Mon-20200706/804078.html)) which mentions:

> It gets miscompiled with GCC 5.3, causing Clang to crash with
"error: unknown target CPU 'x86-64'"

Testing with GCC 10.2 seems to have solved the issue. I'll determine the lowest version I can use.

Note: The commit above references repro

giordano:
I don't understand: isn't that patch in our source code?  Shouldn't it solve the issue with GCC 5?

vchuravy:
Aren't we using musl GCC for bootstrapping?

omus:
> I don't understand: isn't that patch in our source code? Shouldn't it solve the issue with GCC 5?

You are correct that the patch I mentioned is included in LLVM 11.0.1. I believe there is another issue present that hasn't been solved which can be worked around by using a newer GCC.

> Aren't we using musl GCC for bootstrapping?

Originally I thought we were using the system musl GCC for bootstrapping LLVMBootstrap. However, when I encountered an error [stating that GCC 5.1+ was required](https://github.com/JuliaPackaging/Yggdrasil/pull/2473#discussion_r567199650) I used `preferred_gcc_version` to correct the problem which means GCCBootstrap is being used.

omus:
Quick status update:
- Using GCCBoostrap@5.2 can be used to successfully build LLVMBootstrap@11.0.1 but has the `unknown target CPU` under certain environments
- Using GCCBootstrap@6.1 fails to build LLVMBootstrap@11.0.1 at the 50% mark (I didn't look into this further)
- Using GCCBootstrap@7.1 builds LLVMBootstrap@11.0.1 and doesn't have the unknown target error on `x86_64-apple-darwin` or `x86_64-linux-musl`.

omus:
Using new LLVMBootstrap@11.0.1 deployment which should fix the issue we were seeing

giordano:
Note: before merging this I'd like to run the tests in BinaryBuilder as well

giordano:
It looks all good, thank you so much @omus for the great work!

PR #113, opened 2021-02-04
Refactor pigz support

omus:
While building a new GCCBootstrap I noticed that `7z` was being used for compressing the "unpacked" artifact. As the original point of [using `pigz`](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/107) was to improve the compression time when building shards I tracked this down and found this [call to `archive_artifact` in Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil/blob/2fba739eec6d70724d6594a4e5c011fc872cc8e8/0_RootFS/common.jl#L48). In order to fix this I could revise Yggdrasil to use `_archive_artifact` but since this custom version of `archive_artifact` will be used beyond just BinaryBuilderBase I thought it best to clean up the interface first.

Additionally, I used this PR to add support for compressing artifacts in the `xz` format as discussed in https://github.com/JuliaLang/Pkg.jl/pull/2375#issuecomment-771828722. Currently, support for this is opt-in and would be done via:

```julia
archive_artifact(...; package=(src_dir, tarball_path) -> package(src_dir, tarball_path, format="xz"))
```
We could add a convenience function `package(format::AbstractString) = (src_dir, tarball_path) -> package(src_dir, tarball_path, format=format)` but I decided to leave that out for now.

### DIFF
--- a/src/ArchiveUtils.jl
+++ b/src/ArchiveUtils.jl
@@ -151,9 +151,9 @@ end
 
 # Stripped down copy of `Pkg.Artifacts.archive_artifact` that supports a custom `package`
 # function
-function _archive_artifact(hash::SHA1, tarball_path::String;
-                           honor_overrides::Bool=false,
-                           package::Function=_package_fast)
+function archive_artifact(hash::SHA1, tarball_path::String;
+                          honor_overrides::Bool=false,
+                          package::Function=package)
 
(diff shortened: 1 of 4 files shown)


omus:
I'll add some basic tests for this

giordano (approved):
Looks great!

PR #115, opened 2021-02-18
[BuildToolchains] Set `HOST_SYSTEM_NAME` for the host CMake toolchain

### LINKED ISSUE #114: CMake host toolchain file should set `CMAKE_HOST_SYSTEM_NAME`
https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING.html

> This variable will be set to true by CMake if the `CMAKE_SYSTEM_NAME` variable has been set manually (i.e. in a toolchain file or as a cache entry from the `cmake` command line). In most cases, manually setting `CMAKE_SYSTEM_NAME` will only be done when cross compiling, since it will otherwise be given the same value as `CMAKE_HOST_SYSTEM_NAME` if not manually set, which is correct for the non-cross-compiling case. In the event that `CMAKE_SYSTEM_NAME` is manually set to the same value as `CMAKE_HOST_SYSTEM_NAME`, then `CMAKE_CROSSCOMPILING` will still be set to true.

In
https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/de1e65ff11459ae0ed8b8d1318ceea1c0f7c2893/src/BuildToolchains.jl#L42
we always set `CMAKE_SYSTEM_NAME`, so CMake will always think we're doing a cross-compilation, even when we use the toolchain file for the host.

Came out in https://github.com/JuliaPackaging/Yggdrasil/pull/2564

giordano:
Fix #114.

@staticfloat I believe you had a slightly different idea about this, so that
`CMAKE_TARGET_TOOLCHAIN` always sets `SYSTEM_NAME` and `CMAKE_HOST_TOOLCHAIN`
always sets `HOST_SYSTEM_NAME`, but I feel like there is value in always
recognising that we're doing a native build, it makes things easier and
technically it's true that we aren't cross-compiling (even though formally we
are).  What do you think?

### DIFF
--- a/src/BuildToolchains.jl
+++ b/src/BuildToolchains.jl
@@ -29,7 +29,7 @@ function cmake_os(p::AbstractPlatform)
     end
 end
 
-function toolchain_file(bt::CMake, p::AbstractPlatform)
+function toolchain_file(bt::CMake, p::AbstractPlatform; is_host::Bool=false)
     target = triplet(p)
     aatarget = aatriplet(p)
 
@@ -71,9 +71,14 @@ function toolchain_file(bt::CMake, p::AbstractPlatform)
         endif()
         """
     else
+        if is_host
+            system_name_var = "HOST_SYSTEM_NAME"
+        else
+            system_name_var = "SYSTEM_NAME"
+        end
         return """
         # CMake toolchain file for $(c_compiler(bt)) running on $(target)
-        set(CMAKE_SYSTEM_NAME $(cmake_os(p)))
+        set($(system_name_var) $(cmake_os(p)))
         set(CMAKE_SYSTEM_PROCESSOR $(cmake_arch(p)))
 
         set(CMAKE_SYSROOT /opt/$(aatarget)/$(aatarget)/sys-root/)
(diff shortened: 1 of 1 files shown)


staticfloat:
The thing I'm worried about is that now one of the targets is "special"; I can just imagine some recipe being built only for `x86_64-linux-musl` and working, but then failing when we build for `x86_64-linux-gnu` for no other reason than some change in `cmake` behavior which has nothing to do with the actual source code.  I would rather things be consistent across all targets and then only allow users to try and bootstrap things by using `CMAKE_HOST_TOOLCHAIN_FILE`.  This will be useful in the event that we ever decide to change the host triplet; if we decided for some reason to have an `aarch64-linux-gnu` host, it will be a lot easier if we don't have the implicit assumption that if `${target} == x86_64-linux-musl` then you're not cross-compiling; I think it's a lot neater to have _all_ targets be cross-compiling and the only way you're allowed to think you're not cross compiling is to use the `HOST` tools.

giordano:
How about this:
```
% julia-master -e 'using BinaryBuilderBase; BinaryBuilderBase.runshell(Platform("x86_64", "macos"); verbose=false)'
[ Info: Running privileged container via `sudo`, may ask for your password:
sandbox:${WORKSPACE} # ls ${CMAKE_HOST_TOOLCHAIN} 
/opt/toolchains/x86_64-linux-musl-cxx11/host_x86_64-linux-musl.cmake
sandbox:${WORKSPACE} # head -n2 ${CMAKE_HOST_TOOLCHAIN} 
# CMake toolchain file for gcc running on x86_64-linux-musl-cxx11
set(HOST_SYSTEM_NAME Linux)
sandbox:${WORKSPACE} # ls ${CMAKE_TARGET_TOOLCHAIN} 
/opt/toolchains/x86_64-apple-darwin14-libgfortran3-cxx03/target_x86_64-apple-darwin14.cmake
sandbox:${WORKSPACE} # head -n2 ${CMAKE_TARGET_TOOLCHAIN} 
# CMake toolchain file for clang running on x86_64-apple-darwin14-libgfortran3-cxx03
set(SYSTEM_NAME Darwin)
sandbox:${WORKSPACE} # ls /opt/toolchains/*/*.cmake
/opt/toolchains/x86_64-apple-darwin14-libgfortran3-cxx03/target_x86_64-apple-darwin14.cmake
/opt/toolchains/x86_64-apple-darwin14-libgfortran3-cxx03/target_x86_64-apple-darwin14_clang.cmake
/opt/toolchains/x86_64-apple-darwin14-libgfortran3-cxx03/target_x86_64-apple-darwin14_gcc.cmake
/opt/toolchains/x86_64-linux-musl-cxx11/host_x86_64-linux-musl.cmake
/opt/toolchains/x86_64-linux-musl-cxx11/host_x86_64-linux-musl_clang.cmake
/opt/toolchains/x86_64-linux-musl-cxx11/host_x86_64-linux-musl_gcc.cmake
/opt/toolchains/x86_64-linux-musl-cxx11/target_x86_64-linux-musl.cmake
/opt/toolchains/x86_64-linux-musl-cxx11/target_x86_64-linux-musl_clang.cmake
/opt/toolchains/x86_64-linux-musl-cxx11/target_x86_64-linux-musl_gcc.cmake
```
?

PR #117, opened 2021-02-21
Mount existing `/etc/resolv.conf` in `UserNSRunner` if it has nameserver settings

sloede:
As discussed with @staticfloat and @giordano on Slack (https://julialang.slack.com/archives/C674ELDNX/p1613895520011000?thread_ts=1613741887.479600&cid=C674ELDNX).

The purpose is to avoid problems on systems where specific nameserver settings are applied in `/etc/resolv.conf`. In this case it makes sense to use the local `/etc/resolv.conf`, e.g., if the default DNS servers used by `BinaryBuilderBase` (Google's DNS servers) are not reachable from within the local network.

I have tested this locally on an Ubuntu machine and it fixed the issue I had with unreachable Google DNS servers, but I have not been able to test it on non-Ubuntu systems.

@staticfloat Is this implementation what you had in mind? One downside of the current approach is that `/etc/resolv.conf` is not editable anymore, precluding any way for the user to manually fix DNS issues (as I did before). Would it make sense to mount `/etc/resolv.conf` as a writable file? And if yes, how to go about it? I assume one would have to tell Docker (?) to not mount the file but really copy it.

### DIFF
--- a/src/UserNSRunner.jl
+++ b/src/UserNSRunner.jl
@@ -67,14 +67,14 @@ function UserNSRunner(workspace_root::String;
     # specified and if yes, mount the existing `/etc/resolv.conf` to make use
     # of system-specific nameserver settings
     if (Sys.isbsd() || Sys.islinux()) && isfile("/etc/resolv.conf")
-      if any(startswith(line, "nameserver") for line in eachline("/etc/resolv.conf"))
-        # We copy the contents of `/etc/resolv.conf` to a temporary file and mount that one,
-        # such that `/etc/resolv.conf` is editable inside the container
-        tmppath, tmpio = mktemp()
-        write(tmpio, read("/etc/resolv.conf", String))
-        flush(tmpio) # required as otherwise `tmppath` remains empty
-        push!(workspaces, tmppath => "/etc/resolv.conf")
-      end
+        if any(startswith(line, "nameserver") for line in eachline("/etc/resolv.conf"))
+            # We copy the contents of `/etc/resolv.conf` to a temporary file and mount that one,
+            # such that `/etc/resolv.conf` is editable inside the container
+            tmppath, tmpio = mktemp()
+            write(tmpio, read("/etc/resolv.conf", String))
+            flush(tmpio) # required as otherwise `tmppath` remains empty
+            push!(workspaces, tmppath => "/etc/resolv.conf")
+        end
     end
 
     if isnothing(shards)
--- a/src/UserNSRunner.jl
+++ b/src/UserNSRunner.jl
@@ -67,12 +67,9 @@ function UserNSRunner(workspace_root::String;
     # specified and if yes, mount the existing `/etc/resolv.conf` to make use
     # of system-specific nameserver settings
     if (Sys.isbsd() || Sys.islinux()) && isfile("/etc/resolv.conf")
-        for line in eachline("/etc/resolv.conf")
-            if startswith(line, "nameserver")
-              push!(workspaces, "/etc/resolv.conf" => "/etc/resolv.conf")
-              break
-            end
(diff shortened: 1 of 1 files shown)


src/UserNSRunner.jl
```diff
@@ -62,6 +62,19 @@ function UserNSRunner(workspace_root::String;
...
+    # If we are on a system that uses `/etc/resolv.conf` as a resolver
+    # configuration file (mainly *BSD and Linux), check if a nameserver is
+    # specified and if yes, mount the existing `/etc/resolv.conf` to make use
+    # of system-specific nameserver settings
+    if (Sys.isbsd() || Sys.islinux()) && isfile("/etc/resolv.conf")
+        for line in eachline("/etc/resolv.conf")
+            if startswith(line, "nameserver")
+              push!(workspaces, "/etc/resolv.conf" => "/etc/resolv.conf")
+              break
+            end
```
staticfloat, line 74: So this will push an entry onto `workspaces` for _each_ `line` that starts with `nameserver`; not quite what we want.  :)

Try something more like `if any(startswith(line, "nameserver") for line in eachline("/etc/resolv.conf"))`.

staticfloat (approved):

staticfloat:
Yeah, this looks about right!  To maintain editability can you try copying `/etc/resolv.conf` to a temporary file and mounting that file?  That way it should be owned and editable by the current user.

src/UserNSRunner.jl
sloede: I don't think it will do that, does it? The `break` will exit the loop over each lines once the first matching line is found, thus executing `push!` at most once.

Anyhow, since my approach is obviously not as easy to understand, I used your approach in be21700a584cc088bda0b09bf7e9d419b6a30277.

sloede:
>  To maintain editability can you try copying `/etc/resolv.conf` to a temporary file and mounting that file?

I did this in 646804cd8d47cdb3dff4c1afa944d9c9d0087023 and it seems to work on my systems.

src/UserNSRunner.jl
staticfloat: Wow, my eyes totally missed the `break`.  You're right, of course.

src/UserNSRunner.jl
```diff
@@ -62,6 +62,21 @@ function UserNSRunner(workspace_root::String;
...
+    if (Sys.isbsd() || Sys.islinux()) && isfile("/etc/resolv.conf")
+      if any(startswith(line, "nameserver") for line in eachline("/etc/resolv.conf"))
+        # We copy the contents of `/etc/resolv.conf` to a temporary file and mount that one,
+        # such that `/etc/resolv.conf` is editable inside the container
+        tmppath, tmpio = mktemp()
+        write(tmpio, read("/etc/resolv.conf", String))
+        flush(tmpio) # required as otherwise `tmppath` remains empty
+        push!(workspaces, tmppath => "/etc/resolv.conf")
+      end
+    end
```
staticfloat, line 78: this is really nitpicky, but can you indent 4 spaces instead of just 2 to match the rest of the file?  Thanks!

sloede: Sorry, my bad. Fixed in f065e57.

staticfloat:
Thanks @sloede!

giordano:
Any specific reasons why this wasn't done for the Docker runner?

staticfloat:
Docker automatically does this:

```
$ docker run -ti alpine /bin/sh -c "cat /etc/resolv.conf"
# This is Elliot's /etc/resolv.conf
search local
nameserver 1.1.1.1
nameserver 8.8.8.8
nameserver 8.8.4.4
```

PR #118, opened 2021-02-27
[Runner] Correctly use GCC march flags for GCC

giordano:


staticfloat (approved):
How did this ever work?!

vchuravy:
Clangs tried to be GCC compatible

giordano:
The reason is probably because the only different flags are
https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/e1e18b46afc72b515223640a91f8fdc1c0203453/src/Platforms.jl#L123-L140
but we never build for those microarchitectures

PR #98, opened 2021-01-09
Revert "Properly (de)serialize VersionSpec to JSON"

giordano:
Reverts JuliaPackaging/BinaryBuilderBase.jl#93.  It breaks registration of LLVM: https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=8363&view=logs&j=82ede541-d8dc-5eb8-fd6b-5903678e9282&t=31eefa9e-75b0-5107-6058-36e5c2f2fd84
```
    Updating registry at `/depot/registries/General`
Downloading sources...
ERROR: LoadError: KeyError: key "ranges" not found
Stacktrace:
  [1] getindex
    @ ./dict.jl:482 [inlined]
  [2] dejson(#unused#::Type{Pkg.Types.VersionSpec}, d::Dict{String, Any})
    @ BinaryBuilderBase /depot/packages/BinaryBuilderBase/OykWl/src/Dependencies.jl:130
  [3] dependencify(d::Dict{String, Any})
    @ BinaryBuilderBase /depot/packages/BinaryBuilderBase/OykWl/src/Dependencies.jl:138
  [4] _broadcast_getindex_evalf
    @ ./broadcast.jl:648 [inlined]
  [5] _broadcast_getindex
    @ ./broadcast.jl:621 [inlined]
  [6] getindex
    @ ./broadcast.jl:575 [inlined]
  [7] copy(bc::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Tuple{Base.OneTo{Int64}}, typeof(BinaryBuilderBase.dependencify), Tuple{Vector{Any}}})
    @ Base.Broadcast ./broadcast.jl:922
  [8] materialize
    @ ./broadcast.jl:883 [inlined]
  [9] cleanup_merged_object!(meta::Dict{Any, Any})
    @ BinaryBuilder /depot/packages/BinaryBuilder/yQq4Z/src/Declarative.jl:33
 [10] top-level scope
    @ /agent/_work/1/s/.ci/download_sources.jl:14
in expression starting at /agent/_work/1/s/.ci/download_sources.jl:14
```

CC: @fingolfin

PR #123, opened 2021-03-03
Allow storing `compat` strings in Dependency objects

### LINKED ISSUE #89: Meta JSON serialization does not preserve VersionSpec

fingolfin:
... which can then later be put into the generated Project.toml.

This is companion to PR https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1023, and part of a new effort to address https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/89. My previous attempt, https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/98, was not sufficient, as it failed to address the core issue: that while a compat string can be turned into a VersionSpec, the reverse is in general not possible; so specifying the compat bounds using a VersionSpec isn't too useful. The solution chosen here is to instead simply store it as a string (validated by calling `semver_spec` on it).

Resolves #89


src/Dependencies.jl
```diff
@@ -57,13 +58,33 @@ is_runtime_dependency
...
+
+The optional keyword argument `compat` can be used to specify a string for
+use in the `Project.toml` of the generated Julia package.
 """
 struct Dependency <: AbstractDependency
     pkg::PkgSpec
     build_version::Union{VersionNumber,Nothing}
-    Dependency(pkg::PkgSpec, build_version = nothing) = new(pkg, build_version)
+    compat::String  # semver string for use in Project.toml of the JLL
+    function Dependency(pkg::PkgSpec, build_version = nothing; compat::String = "")
```
staticfloat, line 72: I almost wonder if we should just not allow `PkgSpec` objects anymore.  We won't ever look up a package outside of `General`, so I'm not sure what the use of a `PkgSpec` is anymore....

staticfloat: Maybe for backwards-compatibility, we can define a converter method that takes in a `PkgSpec`, extracts just the name, then passes that to the "true" `Dependency()` constructor, which stores only the name.

staticfloat (changes requested):
Overall looks really nice!

src/Dependencies.jl
staticfloat: @giordano do you think we should change `Dependency` to store `name::String` instead of `pkg::PkgSpec`?

fingolfin: Interesting. That would indeed simplify some things (also make the code easier to understand, IMHO).

Naively thought, this seems fine to me. However, one then looses the ability to also specify a UUID. I've never been a fan of that in `build_tarballs.jl` scripts, *but* perhaps there are reasons why being able to do so might be useful? Like, those supply chain attacks some weeks ago against big companies, where attackers registered public packages with names identical to internal only packages, and then internal tooling got confused and picked up those malicious packages instead... That said, if I know the name of an internal package to attack, I might also know its UUID, so I am not sure this would be a valid defense.

I think none of this matter for Yggdrasil, but what about other users?

giordano: In Yggdrasil we won't build packages outside of General, but users do: `PackageSpec` is useful when you want to build using unregistered dependencies (it's even [documented](https://docs.binarybuilder.org/dev/FAQ/#Can-I-publish-a-JLL-package-locally-without-going-through-Yggdrasil?-1) :slightly_smiling_face:) or dependencies deployed locally.  Also, I can think of users using JLLs in their own registries.

fingolfin: So leave in PackageSpec? Fine by me (also: less work ;-))

src/BinaryBuilderBase.jl
```diff
@@ -17,7 +17,7 @@ export AbstractSource, AbstractDependency, SetupSource, PatchSource,
     generate_compiler_wrappers!, preferred_runner, CompilerShard, UserNSRunner,
     DockerRunner, choose_shards, exeext, preferred_libgfortran_version,
     preferred_cxxstring_abi, gcc_version, available_gcc_builds, getversion,
-    getpkg, aatriplet,
+    getpkg, aatriplet, getcompat,
```
fingolfin, line 20: Should this even be exported? I only did it for the test suite, but that could use it qualified (or we could even get rid of it, and just directly access `dep.compat`.

staticfloat: Yeah, let's export it.  We're pretty liberal with what BBB exports since it's mostly just stuff that used to be in BB.  :P

staticfloat (approved):

staticfloat:
I've rebased and pushed a commit which should hopefully fix the test issues.  @fingolfin let me know if this looks reasonable.

src/Dependencies.jl
```diff
@@ -72,10 +72,14 @@ struct Dependency <: AbstractDependency
...
-                error("build_version and compat for $(pkg) are incompatible")
+                throw(ArgumentError("build_version and compat for $(pkg) are incompatible"))
             end
             if pkg.version != Pkg.Types.VersionSpec("*") && !(pkg.version in spec)
-                error("PackageSpec version and compat for $(pkg) are incompatible")
+                throw(ArgumentError("PackageSpec version and compat for $(pkg) are incompatible"))
+            end
+        else
+            if pkg.version != Pkg.Types.VersionSpec("*")
+                throw(ArgumentError("Use the `compat` kwarg to `Dependency` rather than the `version` kwarg to `PackageSpec`"))
```
fingolfin, line 82: I think this will break BB; see the source code comments in my PR therr

staticfloat: I added this to satisfy [this test](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/123/files#diff-7191d1cf5dee1b5b6cdf1dbe4abba72c5880caa2f3f4a8f3d5c0726b07a3259bR50). Should the test be removed then?

fingolfin: Sorry, I left that in by mistake. Should be fixed now.

(And I'd *like* to have that error, but that would require bigger changes in BB and BBB -- the split into two packages doesn't make this easier :/. Perhaps it can be addressed in a future PR, but for now I'd prefer to keep it simple, to reduce the risk of me introducing regressions.)

staticfloat:
Thanks @fingolfin!

PR #125, opened 2021-03-22
expand_cxxstring_abis(Vector{<:AbstractPlatform})

mkitti:
`expand_cxxstring_abis(ps::Vector{<:Platform}; kwargs...)` should be expanded to `expand_cxxstring_abis(ps::Vector{<:AbstractPlatform}; kwargs...)` to match the scalar version of `expand_cxxstring_abis(platform::AbstractPlatform; skip=Sys.isbsd)`


giordano:
This is ok, but it isn't why you were having problems in https://github.com/JuliaPackaging/Yggdrasil/pull/2709

PR #130, opened 2021-04-04
Switch back to Downloads.jl

Wimmerer:
Sounds like this should be ready? Should I add a callback for progress?

### DIFF
--- a/src/ArchiveUtils.jl
+++ b/src/ArchiveUtils.jl
@@ -135,6 +135,8 @@ function download_verify(url, hash, path)
     else
         mkpath(dirname(path))
         @info "Downloading $(url) to $(path)..."
+        
+        rm(path; force=true)
         Downloads.download(url, path)
         verify(path, hash) || error("Verification failed")
     end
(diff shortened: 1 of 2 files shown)


src/ArchiveUtils.jl
```diff
@@ -133,18 +133,7 @@ function download_verify(url, hash, path)
         @info "Cached file found in $(path)"
         return true
     else
-        mkpath(dirname(path))
```
giordano, line 136: I think the mkpath is still necessary

src/ArchiveUtils.jl
```diff
@@ -133,18 +133,7 @@ function download_verify(url, hash, path)
...
-        # `curl` doesn't automatically overwrite an existing file, delete it
-        # before attempting the download.
```
giordano, line 138: We also want to keep the `@info` message and probably also the `rm` (I don't remember whether `Downloads.download` does it automatically)

Wimmerer:
Had to change test file, but I believe that's correct. Do you want me to try and add a progress callback?

giordano:
Thank you so much!

I don't have a strong opinion about progress bars, but in any case it needs to be optional, I don't think we want to on Yggdrasil for example, it'd make the log messy. In any case that's something for a different pull request

PR #133, opened 2021-04-11
[GHA] Run CI also on macOS

giordano:
We need to install Docker with Homebrew as Docker isn't officially supported on
macOS.


DilumAluthge:
```
Error: No machine name(s) specified and no "default" machine exists
Error: Process completed with exit code 1.
```

PR #134, opened 2021-04-14
Fix `clock_gettime` errors on old macOS machines

staticfloat:
Our GCC shards did not correctly configure themselves to target the
oldest supported macOS build.  Instead, because GCC doesn't fully
understand how to use the availability macros that macOS uses, it
assumed that it was able to use `clock_gettime()`, which caused a
reliance on that symbol within `libgfortran` itself.  This resulted in
our CSL being unusable on macOS 10.11 and earlier.  This PR updates the
GCC shards [0] and also adds compiler flags to enforce that users who
attempt to use `gcc` on macOS do not end up making the same mistake.

[0] https://github.com/JuliaPackaging/Yggdrasil/pull/2836

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -236,15 +236,7 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
         return flags
     end
 
-    function min_macos_version_flag(p::AbstractPlatform)
-        # If no `os_version` is specified in `p`, default to the oldest we support in the Julia world,
-        # which is `10.8`, but if it is actually specified, then set that corresponding value.
-        #min_macos_version = something(os_version(p), v"14.0.0")
-
-        # Eventually, we'll take this in `os_version(p)`, but not just yet.  We need to fix the paths
-        # to the compiler shards first, since right now they have `14` at the end
-        min_macos_version = v"14.0.0"
-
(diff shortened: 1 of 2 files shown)


staticfloat:
After this gets merged, we need to rebuild CSL and bump it on base Julia

src/Runner.jl
```diff
@@ -322,6 +331,13 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
             push!(flags, "-D_GLIBCXX_USE_CXX11_ABI=0")
         end
 
+        # Simulate some of the `__OSX_AVAILABLE()` macro usage that is broken in GCC
+        # Currently, we only target 10.10, but eventually, we'll want to tailor this to `os_version(p)`
+        if 14 < 16
```
Keno, line 336: huh?

staticfloat: Eventually, this will be `if os_version(p) < 16`, but since that doesn't work just yet, I'm manually inlining `os_version(p)` as `14`.

giordano (approved):

src/Runner.jl
giordano: I think this is inlining too much, this is set also for non-macOS platforms

PR #135, opened 2021-04-18
Use the proper destdir for cleanup_dependencies

barche:
Attempt to fix the problem seen in https://github.com/JuliaRegistries/General/pull/34597


barche:
I can confirm that this solves the issue with the Qt6Base_jll tarball, but I'm not confident it won't break other stuff, so I won't merge this myself.

giordano (approved):
I feel like `unsymlink_tree` should be a bit smarter in removing the files, but this is probably a good solution for the time being.

PR #136, opened 2021-04-18
Add host_libdir to the LD_LIBRARY_PATH

barche:
This may be needed to run tools included in a HostBuildDependency,
as in Qt6.

See also https://github.com/JuliaPackaging/Yggdrasil/pull/2856#discussion_r615287341

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -917,6 +917,7 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
             csl_paths(host_platform),
             # Add our target/host-specific library directories for compiler support libraries
             target_lib_dir(host_platform),
+            "$host_libdir",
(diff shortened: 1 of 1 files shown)


src/Runner.jl
```diff
@@ -917,6 +917,7 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
             csl_paths(host_platform),
             # Add our target/host-specific library directories for compiler support libraries
             target_lib_dir(host_platform),
+            "$host_libdir",
```
giordano, line 920: I'd just add a comment
```suggestion
            # Libdir of the host platform, to run programs in `HostBuildDependency`
            "$(host_libdir)",
            # Add our target/host-specific library directories for compiler support libraries
            target_lib_dir(host_platform),
```

PR #137, opened 2021-04-20
Bring back old RootFS and make it default

giordano:
Rust toolchain is apparently broken in the new RootFS (I mean, more than usual).
Revert it to previous version until we figure out what's the problem.


giordano:
Running the tests in https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1032, seems to be going well so far.

giordano:
See #139.  Tests in #138 show the error

PR #142, opened 2021-05-01
[CI] Where is Julia being installed?

DilumAluthge:


giordano:
?

DilumAluthge:
Trying to debug https://julialang.slack.com/archives/CPWJ5DGG1/p1619512408105900?thread_ts=1619480226.104400&channel=CPWJ5DGG1&message_ts=1619512408.105900

DilumAluthge:
Basically, I think this means that your CI jobs are not actually using the version of Julia installed by `setup-julia`. I think you're just using whatever default Julia is shipped on the GHA runners. (Because `setup-julia` installs into `/opt`, but you delete `/opt` after running `setup-julia`.)

giordano:
Which is the same version anyway?

DilumAluthge:
Sure, at least right now. But it's confusing, because if someone reads the `jobs` section, they'll think that the versions listed in the `jobs` section are exactly the versions being used by CI, which is not the case.

If you are fine always using the default version of Julia, you could just remove `setup-julia`.

DilumAluthge:
See #143 and #144

PR #143, opened 2021-05-01
[CI] Delete `/opt` before installing Julia (not after)

DilumAluthge:
Closes #144

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,10 +40,7 @@ jobs:
             julia-version: "1.6"
 
     steps:
-      - run: ls -lad /opt
-      - run: ls -lad /opt/hostedtoolcache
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -40,14 +40,16 @@ jobs:
             julia-version: "1.6"
 
     steps:
+      - name: Delete /opt
+        run: sudo rm -rf /opt
+      - name: Create /opt
+        run: sudo mkdir /opt
```
giordano, line 46: Maybe
```suggestion
      - name: Delete /opt
        run: sudo rm -rf /opt/*
```
?

DilumAluthge: Yep that seems to do the trick.

giordano (approved):

PR #129, opened 2021-03-31
Require Julia v1.7

giordano:
It's only one job, but in this way we can catch early failures on nightly

Fix #170, fix https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1130, fix https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1144

PR #144, opened 2021-05-01
[CI] Always run CI on the version of Julia installed by default on GitHub Actions runners

DilumAluthge:
Alternative to #143 

Closes #143


DilumAluthge:
I just saw #129.

If you're planning on eventually doing #129, then this PR won't work.

PR #146, opened 2021-05-16
[Prefix] Add `subdir` argument to `logdir` function

giordano:
Companion pull request in `BinaryBuilder`: https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1059

### DIFF
--- a/src/Prefix.jl
+++ b/src/Prefix.jl
@@ -109,12 +109,13 @@ function includedir(prefix::Prefix)
 end
 
 """
-    logdir(prefix::Prefix)
+    logdir(prefix::Prefix; subdir::AbstractString="")
 
-Returns the logs directory for the given `prefix`.
+Returns the logs directory for the given `prefix`.  If `subdir` is a non-empty string, that
+directory it is appended to the logdir of the given `prefix`.
 """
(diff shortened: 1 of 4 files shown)


src/Prefix.jl
```diff
@@ -109,12 +109,13 @@ function includedir(prefix::Prefix)
...
+    logdir(prefix::Prefix; subdir::AbstractString="")
 
-Returns the logs directory for the given `prefix`.
+Returns the logs directory for the given `prefix`.  If `subdir` is a non-empty string, that
+directory it is appended to the logdir of the given `prefix`.
 """
-function logdir(prefix::Prefix)
-    return joinpath(prefix, "logs")
+function logdir(prefix::Prefix; subdir::AbstractString="")
+    return strip_backslash(joinpath(prefix, "logs", subdir))
```
staticfloat, line 118: Hilariously, `strip_backslash()` actually strips forward-slashes on Linux/macOS.

staticfloat (approved):

src/Prefix.jl
giordano: I think it's a misnomer :facepalm:

PR #148, opened 2021-05-18
[Runner] Improve compile-only flags checks and fix `objc` wrapper

giordano:
The `objc` wrapper is currently roughly equivalent to `clang -x objective-c`,
the problem is that `-x <LANG>` can't be used when linking, otherwise the
compiler driver will assume that the input files are plain text source code
files.

As a simple rule, we assume that we are compiling (as opposed to linking) when
no argument is an object file (ending with `.o` extension) and make the `objc`
wrapper pass the `-x objective-c` flag only when compiling (and not linking),
instead of doing so unconditionally.

Issue spotted while building Glib: https://github.com/JuliaPackaging/Yggdrasil/pull/91#issuecomment-842704490


giordano:
How to reproduce the error:
```console
% julia -e 'using BinaryBuilderBase; BinaryBuilderBase.runshell(Platform("x86_64", "macos"))'
sandbox:${WORKSPACE} # cat > test.m << EOF
> #import <Foundation/Foundation.h>
> void test() {
>     @autoreleasepool {
>         NSLog(@"Hello, World!");
>     }
> }
> EOF
sandbox:${WORKSPACE} # objc -c -o test.o test.m
sandbox:${WORKSPACE} # objc -shared -o test.dylib -framework Foundation test.o
test.o:1:1: error: source file is not valid UTF-8
[...] # watch the input object file being dumped to the screen
fatal error: too many errors emitted, stopping now [-ferror-limit=]
522 warnings and 20 errors generated.

```

src/Runner.jl
```diff
@@ -126,10 +126,11 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
             println(io)
         end
 
-        # If we're given compile-only flags, include them only if `-x assembler` is not provided
+        # If we're given compile-only flags, include them only if `-x assembler` is not provided,
+        # or no object files (*.o) are passed in input
         if !isempty(compile_only_flags)
             println(io)
-            println(io, "if [[ \" \${ARGS[@]} \" != *' -x assembler '* ]]; then")
+            println(io, "if [[ \" \${ARGS[@]} \" != *' -x assembler '* ]] && [[ \" \${ARGS[@]} \" != *'.o '* ]]; then")
```
staticfloat, line 133: This seems like a hard problem to solve.  I'm honestly not sure what to do here.

On the one hand, we need a way to tell `clang` that it should be in `-x objc` mode, but it throws an error if we're linking.....  On modern macs, the `objc` executable doesn't even exist, so perhaps this whole compiler wrapper is just a bad idea.  What software are you building that needs this?

giordano: Meson needs it...

giordano: So, this is the problem: we need to specify an Objective-C compiler in the Meson cross-file:
https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/cd3f4d129c4fe5126b13a57979a30c927ae2f0a0/src/BuildToolchains.jl#L167
Meson is very strict in many ways, I'm 99% sure it won't accept something like `clang -x objective-c`, and anyways it uses the compiler also as linker (which is what's happening in https://github.com/JuliaPackaging/Yggdrasil/pull/91#issuecomment-842704490).

giordano: Uhm, let me see if I can do it with
```python
objc = '/opt/bin/$(target)/$(aatarget)-clang'
# ...
objc_args = ['-x objective-c']
```

PR #149, opened 2021-05-19
Reorganise Objective-C wrapper and Meson toolchain file

giordano:
We can't make the Objective-C compiler always be the same as
`clang -x objective-c` because this won't work when using the compiler as linker: when
specifying the language with `-x <LANG>` the input files are assumed to always
be plain text source code files, but you pass object files to the linker.

This changeset removes the `objc` wrapper, which was created solely with the
goal to build Glib with Meson (which we never managed to do so far anyways) and
instead sets in the Meson toolchain file the Objective-C compiler to `cc` and
passes `-x objective-c` only when compiling but not linking.  This is more
correct and less error-prone than what we have now.

We also automatically add `-I${includedir}` and `-L{libdir}` for all languages
in the Meson toolchain file, to avoid having to manually patch this file during
builds, as Meson doesn't read environment variables when doing
cross-compilation.

---

In the previous episodes of the Objective-C compiler wrapper in BinaryBuilder:

* https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/426
* https://github.com/JuliaIO/FFMPEGBuilder/pull/5
* https://github.com/JuliaPackaging/BinaryBuilder.jl/commit/7106456295f34a1e7d791b4b2af8f28a41dd260a
* https://github.com/JuliaPackaging/Yggdrasil/pull/73

Close #148

### DIFF
--- a/src/BuildToolchains.jl
+++ b/src/BuildToolchains.jl
@@ -108,11 +108,17 @@ function toolchain_file(bt::CMake, p::AbstractPlatform; is_host::Bool=false)
     end
 end
 
+meson_c_args(p::AbstractPlatform) = ["'-I/workspace/destdir/include'"]
+meson_cxx_args(p::AbstractPlatform) = meson_c_args(p)
+meson_objc_args(p::AbstractPlatform) = push!(meson_c_args(p), "'-x'", "'objective-c'")
+meson_fortran_args(p::AbstractPlatform) = meson_c_args(p)
+
 function meson_c_link_args(p::AbstractPlatform)
+    libdir = "/workspace/destdir/" * (Sys.iswindows(p) ? "bin" : "lib")
     if arch(p) == "powerpc64le" && Sys.islinux(p)
-        return "'-Wl,-rpath-link,/workspace/destdir/lib64'"
+        return ["'-L$(libdir)'", "'-Wl,-rpath-link,/workspace/destdir/lib64'"]
     else
-        return ""
+        return ["'-L$(libdir)'"]
     end
 end
 meson_cxx_link_args(p::AbstractPlatform) = meson_c_link_args(p)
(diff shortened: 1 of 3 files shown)


src/Runner.jl
```diff
@@ -939,7 +935,7 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
         # Default mappings for some tools
         "CC" => "cc",
         "CXX" => "c++",
-        "OBJC" => "objc",
+        "OBJC" => "cc -x objective-c",
```
giordano, line 938: This is the only thing I'm not 100% sure about: it looks like the `OBJC` environment variable would be useful for FFMPEG (see https://github.com/JuliaIO/FFMPEGBuilder/pull/5 and https://github.com/JuliaPackaging/Yggdrasil/blob/8be19820afa24ff817185f3c53c4d20786a8add3/F/FFMPEG/common.jl#L77), but I'm not sure it's used anywhere else, so it may not be worth keeping it.

src/BuildToolchains.jl
```diff
@@ -108,11 +108,17 @@ function toolchain_file(bt::CMake, p::AbstractPlatform; is_host::Bool=false)
     end
 end
 
+meson_c_args(p::AbstractPlatform) = "'-I/workspace/destdir/include'"
+meson_cxx_args(p::AbstractPlatform) = meson_c_args(p)
+meson_objc_args(p::AbstractPlatform) = meson_c_args(p) * ", '-x', 'objective-c'"
+meson_fortran_args(p::AbstractPlatform) = meson_c_args(p)
```
staticfloat, line 114: Can we make these return arrays, then join them at the end?  I dislike string manipulation like this as it's not very flexible.

src/Runner.jl
staticfloat: FFMPEG uses `OBJCC`, not `OBJC` I think; so my guess is that we're fine just dropping this environment variable completely.

staticfloat (approved):

src/Runner.jl
giordano: Strictly speaking, FFMPEG doesn't use any environment variable, but it may be convenient to have one just in case.  In https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/426#issuecomment-526486197 I observed that I couldn't find a really standard environment variable for Objective-C, people seem to use different alternatives, even though `OBJC` seemed the slightly more common one.

staticfloat: Yeah, I think it's probably better to just not have an environment variable at all then.

src/Runner.jl
```diff
@@ -499,7 +497,7 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
             "FC"     => "$(host_target)-f77",
             "LD"     => "$(host_target)-ld",
             "NM"     => "$(host_target)-nm",
-            "OBJC"   => "$(host_target)-objc",
+            "OBJC"   => "$(host_target)-cc",
```
giordano, line 500: The host Objective-C compiler is _**required**_ by Meson (:facepalm: but I think in newer versions this should be fixed), but it isn't a big deal as this variable is set only inside the `meson` wrapper.

PR #153, opened 2021-06-19
[Rootfs] Add GCC 11.1

giordano:
Companion pull request in Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/pull/3174

### DIFF
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -382,6 +382,7 @@ const available_gcc_builds = [
     GCCBuild(v"8.1.0", (libgfortran_version = v"5", libstdcxx_version = v"3.4.25", cxxstring_abi = "cxx11")),
     GCCBuild(v"9.1.0", (libgfortran_version = v"5", libstdcxx_version = v"3.4.26", cxxstring_abi = "cxx11")),
     GCCBuild(v"10.2.0", (libgfortran_version = v"5", libstdcxx_version = v"3.4.28", cxxstring_abi = "cxx11")),
+    GCCBuild(v"11.1.0", (libgfortran_version = v"5", libstdcxx_version = v"3.4.29", cxxstring_abi = "cxx11")),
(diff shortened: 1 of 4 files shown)


src/Rootfs.jl
```diff
@@ -512,7 +513,7 @@ function choose_shards(p::AbstractPlatform;
     function find_shard(name, version, archive_type; target = nothing)
         # Ugly hack alert!  Because GCC 11 has somehow broken C++, we pair GCC 10 with GCC 11 on MacOS
         if name == "GCCBootstrap" && version.major == 11 && target !== nothing && Sys.islinux(target)
-            version = v"10.2.0"
+            version = v"11.1.0"
```
giordano, line 516: I'm not really sure about this, I've never got the point of this `if` --- I think it was related to the aarch64-apple-darwin GCC?

giordano:
I ran `make -j${nproc} -sC /usr/share/testsuite install` for a bunch of platforms (in particular all x86_64 systems), it seems to be all good.  This should be ready to go.

src/Rootfs.jl
staticfloat: I believe what happened here was that I tried building a GCC 11 build targeting `x86_64-linux-musl`, and then when I targeted `aarch64-apple-darwin`, I would get GCC 11 for both target and host.  Then, when I tried to build stuff targeting the host (e.g. using `HOST_CXX`) everything was broken with GCC 11.  So I made this `if` statement to force it to give us GCC 11 for `aarch64`, but GCC 10 for `x86_64`.

Since you say you were able to run the test suite, it sounds like it's fixed now, so I think we can remove the `if` statement completely.  Thanks!

giordano: Ok, it looks like this is needed to make the aarch64-apple-darwin shell work, otherwise this function wouldn't find a GCCBootstrap with the same version number for the host system.  I suspect the comment is just completely misleading?

staticfloat: Mmm, I think it was actually needed for the reason the comment stated, but the new form looks good to me.  Go ahead and update the comment to reflect the new nature of the conditional, and I think we're good to go.

PR #154, opened 2021-06-23
[Prefix] Do not show warning if clashing files are the same

giordano:
A way to test this is to have the following dependencies:
```julia
dependencies = [
    Dependency("Libuuid_jll", v"2.34"),
    Dependency("Libmount_jll", v"2.34"),
]
```
The two artifacts have dozens of clashing files (all `share/locale/*/LC_MESSAGES/util-linux.mo`), but for the same version of the packages they are the same.

### DIFF
--- a/src/Prefix.jl
+++ b/src/Prefix.jl
@@ -213,6 +213,16 @@ function symlink_tree(src::AbstractString, dest::AbstractString)
             src_file = joinpath(root, f)
             dest_file = joinpath(dest, relpath(root, src), f)
             if isfile(dest_file)
+                # Ugh, destination file already exists.  If source and destination files
+                # have the same size and SHA256 hash, just move on, otherwise issue a
+                # warning.
+                if filesize(src_file) == filesize(dest_file)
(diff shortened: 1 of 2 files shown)


src/Prefix.jl
```diff
@@ -213,6 +213,12 @@ function symlink_tree(src::AbstractString, dest::AbstractString)
             src_file = joinpath(root, f)
             dest_file = joinpath(dest, relpath(root, src), f)
             if isfile(dest_file)
+                src_file_hash = open(io -> bytes2hex(sha256(io)), src_file, "r")
```
staticfloat, line 216: Can we have a quick `if filesize(dest_file) != filesize(src_file)` check?  That should prevent us from having to hash in the majority of cases, and should be much faster.

staticfloat (changes requested):

src/Prefix.jl
giordano: Done!  And I've also added a test about this.

PR #156, opened 2021-06-29
[Runner] Fix value of `LD` environment variable for macOS

giordano:


src/Runner.jl
```diff
@@ -979,7 +979,7 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
 
     # If we're on macOS, we give a hint to things like `configure` that they should use this as the linker
     if Sys.isapple(platform)
-        mapping["LD"] = "/opt/$(target)/bin/ld64.macos"
```
giordano, line 982: @staticfloat why was this set to this value?  There is no such file in our environment:
```console
sandbox:${WORKSPACE} # qfind /opt -name "ld64*"
/opt/bin/x86_64-apple-darwin14-libgfortran3-cxx03/ld64.x86_64-apple-darwin14
```

staticfloat: Hmmmm...... looks like I stopped doing this in https://github.com/JuliaPackaging/Yggdrasil/commit/d787aa20aa7b4807a6aae9eff9f9cf9651b88144#diff-37a7da7ede42eaf0bfe29fb575ba5a7450bbc8a3621dd2a9d758fb04c749d07cL648, but I don't remember what allowed me to get rid of it.  In any case, if this passes the shard testsuite, I'm okay with it.

staticfloat (approved):

src/Runner.jl
giordano: I don't think the macOS is used anywhere in our tests, since at the moment this is broken:
```console
sandbox:${WORKSPACE} # echo ${LD}
/opt/x86_64-apple-darwin14/bin/ld64.macos
sandbox:${WORKSPACE} # ls ${LD}
ls: /opt/x86_64-apple-darwin14/bin/ld64.macos: No such file or directory
```
:slightly_smiling_face:

PR #157, opened 2021-06-29
[Runner] Move `csl_paths(host_platform)` at the beginning of `LD_LIBRARY_PATH`

giordano:
This should make it easier to run applications we build with the host toolchain.  CC: @vchuravy

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -69,12 +69,8 @@ function ld_library_path(target::AbstractPlatform,
     # If requested, start with our CSL libraries for target/host, but only for architectures
     # that can natively run within this environment
     if csl_paths
-        # Ok, this is incredibly finicky.  If the target has the same architecture as the
-        # host, we should have the host first then the target, otherwise we need to have
-        # target first.
-        platforms = arch(target) == arch(host) ? (host, target) : (target, host)
         append!(paths,
-                unique("/usr/lib/csl-$(libc(p))-$(arch(p))" for p in platforms if Sys.islinux(p) && proc_family(p) == "intel"),
+                unique("/usr/lib/csl-$(libc(p))-$(arch(p))" for p in (host, target) if Sys.islinux(p) && proc_family(p) == "intel"),
                 )
     end
 
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -47,6 +47,67 @@ end
 # XXX: we want AnyPlatform to look like `x86_64-linux-musl` in the build environment.
 aatriplet(p::AnyPlatform) = aatriplet(default_host_platform)
 
+function ld_library_path(target::AbstractPlatform,
+                         host::AbstractPlatform,
+                         prefix::String="",
+                         host_libdir::String="";
+                         csl_paths::Bool=true)
+    # Helper for generating the library include path for a target.  MacOS, as usual,
+    # puts things in slightly different place.
+    function target_lib_dir(p::AbstractPlatform)
+        t = aatriplet(p)
+        if Sys.isapple(p)
+            return "/opt/$(t)/$(t)/lib:/opt/$(t)/lib"
+        else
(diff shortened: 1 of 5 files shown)


src/Runner.jl
```diff
@@ -918,12 +918,12 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
         ), ":"),
 
         "LD_LIBRARY_PATH" => join((
-           # Start with a default path
-           "/usr/local/lib64:/usr/local/lib:/usr/lib64:/usr/lib",
+            # Start with our CSL libraries for all architectures that can natively run within this environment
+            csl_paths(host_platform),
+            # Then add default system paths
+            "/usr/local/lib64:/usr/local/lib:/usr/lib64:/usr/lib",
```
giordano, line 924: Honestly, I'm not 100% this is a good idea.  Currently we have:
```console
sandbox:${WORKSPACE} # /usr/bin/gcc --version
gcc (Alpine 9.3.0) 9.3.0
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
Maybe we should move our CSL forward only if our host GCCBootstrap shard is > 9.3?

staticfloat: Hmmmm.  So the struggle here is that we want to pick up e.g. `libgcc_s` that matches the host compiler?

We shouldn't really be using `/usr/bin/gcc` at all, right?  Unless we're bootstrapping....

giordano: No, but the problem is that in https://github.com/JuliaPackaging/Yggdrasil/pull/3221#issuecomment-869701371 system libstdc++.so was picked, when we were building with our GCC 11, so that running an application would result in the now infamous undefined symbol `_ZSt28__throw_bad_array_new_lengthv@@GLIBCXX_3.4.29`

vchuravy:
This keeps blocking new LLVM versions being built for Julia so it would be great to have a solution that allows running binaries built for the host.

giordano:
Tested in https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1032, unless I'm missing something, all failures are in Rust tests... Rust toolchain is hopelessly broken... :cry:

giordano:
This seems to cause some issues: https://github.com/JuliaPackaging/Yggdrasil/pull/3249#issuecomment-879257918  I wonder whether we should override `LD_LIBRARY_PATH` inside the compiler wrappers

giordano:
Ok, this pull request is turning into a massive yak shaving.

Now I'm overriding `LD_LIBRARY_PATH` inside the compiler wrappers and I added a test based on the reproducer above.  However, now I can't run the compiled program for `i686-linux-gnu-libgfortran5`.  Note: this works on `master`, more specifically it works without [`86114d6` (#157)](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/157/commits/86114d6c61f4f7bcfbc27a7e9a5e5541028438fa), this means that running the program only worked because the CSL paths have always been wrong.  Running the test program results in the following error:
```
./test: error while loading shared libraries: libc.musl-i386.so.1: cannot open shared object file: No such file or directory
```
@staticfloat does this mean we aren't rejecting musl correctly?  Any hints?

vchuravy:
>  However, now I can't run the compiled program for i686-linux-gnu-libgfortran5. 

That sounds like a feature xD

giordano:
Ok, I think I found a solution: generate CSL paths only for target and host platforms, at the moment we were generating them for _all_ platforms, resulting in the error shown above: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/157#issuecomment-879426510.  The annoying thing is that the order is very important: target first only if it has an architecture different from the host, otherwise host first.

staticfloat (approved):

PR #165, opened 2021-07-12
[Rootfs] Update to new build

giordano:
Companion pull request in Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/pull/3321.

This is needed for #157, because we need a much newer CSL to (hopefully) make things work.

Note: I didn't upgrade the base image to Alpine v3.13 or v3.14, any opinions
about that?


staticfloat:
> Note: I didn't upgrade the base image to Alpine v3.13 or v3.14, any opinions
about that?

It's nice to have a version number to something that is easily human understood, but beyond that no real opinion.

staticfloat (approved):

giordano:
I guess I should have checked the result before opening the pull request, but the CSL directories seem to have the wrong versions of libstdc++ :confused:

PR #166, opened 2021-08-10
Add PlatformSupport and GCCBootstrap artifacts for FreeBSD 12

ararslan:
Generated from https://github.com/JuliaPackaging/Yggdrasil/pull/3447.

PR #167, opened 2021-08-11
Set version to 1.0.0

ararslan:
As far as I know, #166 is technically a breaking change, as it removes FreeBSD 11 artifacts. Thus we would need to tag a breaking release anyway, but if we're going to do that, we'd might as well tag 1.0 given the general stability of BBB. Thoughts?


giordano (approved):
🚀

staticfloat:
zoom zoom

PR #169, opened 2021-08-17
[BuildToolchains] Teach CMake that libraries go in `bin/` on Windows

giordano:
Hopefully we'll see fewer

```
Simple buildsystem detected; Moving all `.dll` files to `bin`!
```

warnings with CMake-based project, as long as people don't decide to override
the destination directory.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "BinaryBuilderBase"
 uuid = "7f725544-6523-48cd-82d1-3fa08ff4056e"
 authors = ["Elliot Saba <staticfloat@gmail.com>"]
-version = "1.0.0"
+version = "1.0.1"
 
 [deps]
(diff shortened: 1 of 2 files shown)


staticfloat (approved):

giordano:
I'm moderately sure this PR does the right thing, but I can't find not even one package to test with, because apparently no one use `CMAKE_INSTALL_LIBDIR`

giordano:
My understanding is that `CMAKE_INSTALL_LIBDIR` is part of the [`GNUInstallDirs`](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) (which isn't used by many packages), and I believe that does already the right thing: I checked for example in libssh2, which already installs libraries for Windows in `${prefix}/bin` without this PR.

I'm going to close this, unless there is more evidence this is actually useful in some other cases.

PR #171, opened 2021-10-06
add installed_shards() function to output downloaded artifacts on system

Crghilardi:
This PR adds a function `installed_shards()` that outputs a vector of `CompilerShard`s that are currently download and installed on a user's system using the stdlib `Artifacts.artifact_exists()`.

I ran into two situations where this could be useful
1. I downloaded many OS/CompilerVersion shards doing one-off testing and filled up my hard drive. Having this, I would be able to reference which shards I know I use consistently vs ones that could be deleted (still have to manually search and delete).
2. The [Artifact.toml](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/commit/46437bea72d02e5186d16f215b125cd004f52c84) for BBB gets updated and I am trying to see if I have an old version downloaded or the newest version.

Things this can't do:
If the BBB `Artifact.toml` is updated, this won't know anything about the older versions anymore, so those shards would stay downloaded but not show up anymore (freebsd v11 in the link above).

I wasn't sure if this was worth exporting, but `BinaryBuilderBase.installed_shards()` is easy enough to type.
~I don't know if/how this should be tested, so marking as draft for now?~ Other non-exported functions don't seem to be in tests either.

### DIFF
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -970,8 +970,7 @@ end
 Return a vector of compiler shards currently downloaded on the local system
 """
 function installed_shards()
-    hashes = shard_source_artifact_hash.(all_compiler_shards())
-    idx = artifact_exists.(hashes)
+    idx = artifact_exists.(shard_source_artifact_hash.(all_compiler_shards()))
 
     return all_compiler_shards()[idx]
 end
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -964,3 +964,14 @@ function unmount_shards(ur::Runner; verbose::Bool = false)
     catch
     end
 end
+
+"""
+    installed_shards()
+Return a vector of compiler shards currently downloaded on the local system
+"""
+function installed_shards()
+    hashes = shard_source_artifact_hash.(all_compiler_shards())
+    idx = artifact_exists.(hashes)
+
+    return all_compiler_shards()[idx]
+end


giordano:
Hi, I somewhat missed this PR so far (probably because I was still ignoring the repo during my holidays)!

The function you add here is interesting and the implementation looks good, but I feel like it isn't needed that often (I never needed it, for once :sweat_smile:).  So I'm not sure about adding a dead code path that may even get removed in the future because it's unused.

Crghilardi:
That's funny, because I feel like I have wanted this function a lot (mainly due to point 1 above) :laughing:

Maybe this makes more sense with what I was hoping to develop on top of this?
What I was imagining with this snippet of functionality is some kind of `manage_compilers()` function that allows a `y/n` type prompt to allow for finer level of control of install/uninstall of compiler shard artifacts. You need to know which artifacts have currently been installed to be able to remove them.

poor artwork mock up:

```
i = install, d = delete, q = done
[] CompilerShard(MacOS, ....)
[] CompilerShard(MacOS, ....)
[] CompilerShard(linux, musl, ....)

```

Is there an opportunity for a middle ground between CompilerShards being lazily downloaded artifacts and [downloading everything](https://juliapackaging.github.io/BinaryBuilderBase.jl/stable/#BinaryBuilderBase.download_all_artifacts-Tuple{}) or is the answer always to get a larger hard drive?

src/Rootfs.jl
```diff
@@ -964,3 +964,14 @@ function unmount_shards(ur::Runner; verbose::Bool = false)
...
     end
 end
+
+"""
+    installed_shards()
+Return a vector of compiler shards currently downloaded on the local system
+"""
+function installed_shards()
+    hashes = shard_source_artifact_hash.(all_compiler_shards())
+    idx = artifact_exists.(hashes)
```
giordano, line 974: Minor micro-optimisation (fuse all broadcasting and avoid a temporary array):
```suggestion
    idx = artifact_exists.(shard_source_artifact_hash.(all_compiler_shards()))
```

giordano (approved):
Alright, let's do it, I just wanted to use something to check what version of GCC I have for the aarch64 Linux platforms:
```julia
julia> sort(filter(s -> s.name == "GCCBootstrap" && Sys.islinux(s.target) && arch(s.target) == "aarch64", installed_shards()); by=x->x.version)
14-element Vector{CompilerShard}:
 CompilerShard("GCCBootstrap", v"4.8.5", Linux aarch64 {libc=musl}, Linux x86_64 {libc=musl}, :unpacked)
 CompilerShard("GCCBootstrap", v"4.8.5", Linux aarch64 {libc=glibc}, Linux x86_64 {libc=musl}, :unpacked)
 CompilerShard("GCCBootstrap", v"4.8.5", Linux aarch64 {libc=musl}, Linux x86_64 {libc=musl}, :squashfs)
 CompilerShard("GCCBootstrap", v"4.8.5", Linux aarch64 {libc=glibc}, Linux x86_64 {libc=musl}, :squashfs)
 CompilerShard("GCCBootstrap", v"5.2.0", Linux aarch64 {libc=glibc}, Linux x86_64 {libc=musl}, :squashfs)
 CompilerShard("GCCBootstrap", v"5.2.0", Linux aarch64 {libc=musl}, Linux x86_64 {libc=musl}, :squashfs)
 CompilerShard("GCCBootstrap", v"6.1.0", Linux aarch64 {libc=musl}, Linux x86_64 {libc=musl}, :squashfs)
 CompilerShard("GCCBootstrap", v"6.1.0", Linux aarch64 {libc=glibc}, Linux x86_64 {libc=musl}, :squashfs)
 CompilerShard("GCCBootstrap", v"7.1.0", Linux aarch64 {libc=glibc}, Linux x86_64 {libc=musl}, :unpacked)
 CompilerShard("GCCBootstrap", v"7.1.0", Linux aarch64 {libc=musl}, Linux x86_64 {libc=musl}, :unpacked)
 CompilerShard("GCCBootstrap", v"7.1.0", Linux aarch64 {libc=glibc}, Linux x86_64 {libc=musl}, :squashfs)
 CompilerShard("GCCBootstrap", v"7.1.0", Linux aarch64 {libc=musl}, Linux x86_64 {libc=musl}, :squashfs)
 CompilerShard("GCCBootstrap", v"8.1.0", Linux aarch64 {libc=glibc}, Linux x86_64 {libc=musl}, :squashfs)
 CompilerShard("GCCBootstrap", v"8.1.0", Linux aarch64 {libc=musl}, Linux x86_64 {libc=musl}, :squashfs)
```
:slightly_smiling_face:

PR #173, opened 2021-10-17
Clarify error messages in compiler wrappers

marcom:
Mention illegal compiler flags in error message to make it clearer what the problem is


giordano:
Thanks!  Note that you have to update the tests accordingly

Issue #174, opened 2021-10-18
[Dependency] Add a field to filter platforms where a dependency is used


giordano:
I think it could be useful to have a field to the `Dependency` struct to filter the platforms where the package will actually be loaded.  We can't have platform-dependent dependencies because Pkg doesn't allow it, but we can at least avoid loading them if not necessary, even if they are mostly no-op.  For example, GTK3_jll on macOS and Windows could not load the whole X11 stack, and NativeFileDialog_jll could not load GTK3_jll on the same platforms.

CC: @fingolfin

giordano:
Fixed by #176 and https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1129

PR #176, opened 2021-11-03
Add field to dependencies to filter them by platform

giordano:
This let us declare that certain dependencies are needed at build/runtime only
on some platforms.

First part to address #174, next part will be in BinaryBuilder (I have a somewhat working version locally, but need to figure out a couple of things not working as expected).  Opening as draft for the time being to ask for feedback, especially about names (I'm not really fond of the name `filter_platforms`, which is also used for both the field and the function to filter the vector of dependencies).  After settling on names and API I'll add documentation and tests.

You can find a first proof-of-concept at https://github.com/giordano/NativeFileDialog_jll.jl (note: I'll probably delete the repo after this work is done).  You can see that GTK3_jll is in the [project](https://github.com/giordano/NativeFileDialog_jll.jl/blob/6ade8b1d2155ad93a3a61b242980c267b6afe927/Project.toml) because we can't have conditional dependencies, but it isn't loaded for the [macOS wrapper](https://github.com/giordano/NativeFileDialog_jll.jl/blob/6ade8b1d2155ad93a3a61b242980c267b6afe927/src/wrappers/aarch64-apple-darwin.jl), while it is for the [Linux one](https://github.com/giordano/NativeFileDialog_jll.jl/blob/6ade8b1d2155ad93a3a61b242980c267b6afe927/src/wrappers/x86_64-linux-gnu.jl).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "BinaryBuilderBase"
 uuid = "7f725544-6523-48cd-82d1-3fa08ff4056e"
 authors = ["Elliot Saba <staticfloat@gmail.com>"]
-version = "1.0.4"
+version = "1.1.0"
 
 [deps]
 CodecZlib = "944b1d66-785c-5afd-91f1-9de20f533193"
--- a/src/Dependencies.jl
+++ b/src/Dependencies.jl
@@ -1,8 +1,7 @@
 using UUIDs
 
 export Dependency, BuildDependency, HostBuildDependency,
-    is_host_dependency, is_target_dependency, is_build_dependency, is_runtime_dependency,
-    filter_platforms
+    is_host_dependency, is_target_dependency, is_build_dependency, is_runtime_dependency
 
 
 # Pkg.PackageSpec return different types in different Julia versions so...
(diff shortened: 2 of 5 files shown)


staticfloat (approved):
Is it normal for us to expect the `Function` to be of the form `platforms -> filter(..., platforms)`, or just to be the `...`?  Basically I'm asking, should we be expecting the user to pass in something like `p -> !Sys.iswindows(p)`, or to pass in something like `ps -> filter(p -> !Sys.iswindows(p), ps)`?

giordano:
I guess showing the code is easier: https://github.com/giordano/BinaryBuilder.jl/commit/f83156ecc3a002ffcd6b5a5c9d20513002b67dc4.  In BB we always have situations where we loop over the platforms (e.g. when building the tarball, or generating the JLL) and have a vector of dependencies, that's why I went with a function of a single platform.  Then there is the `filter_platforms(deps::AbstractVector{<:AbstractDependency}, p::AbstractPlatform)` method which returns the list of dependencies which are requested for the given platform.

fingolfin (approved):
Great!!

vchuravy (approved):
Fantastic!

giordano:
Ok, there is still something funny I need to work out in my [patch to BinaryBuilder](https://github.com/JuliaPackaging/BinaryBuilder.jl/compare/master...giordano:mg/deps-filter-platforms), but I just realised there is a major problem: we need to serialise/deserialise the dependencies to/from JSON, how the heck this is going to work? :sob:

giordano:
Allright, I believe this is now ready for full review.

I changed the new field from a function (hard to serialise with JSON, which I think is a property we want to preserve) to a vector of platforms, easier to handle in that case.  However, I still believe a function would be better, having a vector of platforms is error-prone, since in principle there is no relation between the platforms and the dependencies in a build recipe, which may accidentally get out-of-sync.

vchuravy (approved):

Issue #177, opened 2021-11-05
Allow non-x86_64 host architectures


vchuravy:
Poking about at https://github.com/JuliaCI/PkgEval.jl/issues/97 today I got to the step where we instantiate a build environment using BinaryBuilder.

```
[ Info: Running privileged container via `sudo`, may ask for your password:
 Downloading artifact: PlatformSupport-powerpc64le-linux-gnu.v2021.1.28.x86_64-linux-musl.unpacked
  Downloaded artifact: PlatformSupport-powerpc64le-linux-gnu.v2021.1.28.x86_64-linux-musl.unpacked
 Downloading artifact: GCCBootstrap-powerpc64le-linux-gnu.v7.1.0.x86_64-linux-musl.unpacked
  Downloaded artifact: GCCBootstrap-powerpc64le-linux-gnu.v7.1.0.x86_64-linux-musl.unpacked
 Downloading artifact: LLVMBootstrap.v11.0.1.x86_64-linux-musl.unpacked
  Downloaded artifact: LLVMBootstrap.v11.0.1.x86_64-linux-musl.unpacked
 Downloading artifact: PlatformSupport-x86_64-linux-musl.v2021.1.28.x86_64-linux-musl.unpacked
  Downloaded artifact: PlatformSupport-x86_64-linux-musl.v2021.1.28.x86_64-linux-musl.unpacked
 Downloading artifact: GCCBootstrap-x86_64-linux-musl.v7.1.0.x86_64-linux-musl.unpacked
    Downloading [>                                        ]  0.0 %
```

Would be great if we could support more than just x86_64 as a host arch.

giordano:
😂 😂 😂

staticfloat:
I've made some small progress toward this in my GCC rework branch, but I still consider this pie-in-the-sky.

PR #178, opened 2021-11-06
Shake out some assumptions about architecture of host platform

giordano:
Ref #177.


giordano:
The error
```
  Unable to automatically install 'GCCBootstrap-aarch64-linux-gnu.v4.8.5.x86_64-linux-musl.unpacked' from '/home/runner/work/BinaryBuilderBase.jl/BinaryBuilderBase.jl/Artifacts.toml'
```
is unrelated and we're seeing it everywhere, I'm not really sure what's going on.

vchuravy (approved):
Thanks! This looks like a step in the right direction

PR #183, opened 2021-11-30
Clear out artifacts destdir path if it already exists

Keno:
Since we're doing the symlinking now, there's no reason not to just
have BinaryBuilderBase manage this directory and clear it out if it
already exists. This avoids having to manually clear out existing
directories in `bb add`.


src/Prefix.jl
```diff
@@ -512,6 +512,9 @@ function setup_dependencies(prefix::Prefix, dependencies::Vector{PkgSpec}, platf
             # Copy the artifact from the global installation location into this build-specific artifacts collection
             src_path = Pkg.Artifacts.artifact_path(Base.SHA1(meta["git-tree-sha1"]))
             dest_path = joinpath(prefix, triplet(platform), "artifacts", basename(src_path))
+            if isdir(dest_path)
+                rm(dest_path; recursive=true)
+            end
```
giordano, line 517: Might just do
```suggestion
            rm(dest_path; force=true, recursive=true)
```

giordano (approved):
Looks good to me, besides the fact we could just use `force=true` instead of the `if`

PR #184, opened 2021-12-01
Upgrade Rust to v1.57.0 and JIT-generate Cargo config file

giordano:
This will replace the file we currently hard-code in the Rust compiler shard.

PR to update Rust toolchain build script: https://github.com/JuliaPackaging/Yggdrasil/pull/3992

### DIFF
--- a/src/BuildToolchains.jl
+++ b/src/BuildToolchains.jl
@@ -214,8 +214,8 @@ function toolchain_file(bt::Meson, p::AbstractPlatform, envs::Dict{String,String
     """
 end
 
-function generate_toolchain_files!(platform::AbstractPlatform, envs::Dict{String,String};
-                                   toolchains_path::AbstractString,
+function generate_toolchain_files!(platform::AbstractPlatform, envs::Dict{String,String},
+                                   toolchains_path::AbstractString;
                                    host_platform::AbstractPlatform = default_host_platform,
                                    )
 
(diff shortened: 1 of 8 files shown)


src/DockerRunner.jl
```diff
@@ -93,9 +93,18 @@ function DockerRunner(workspace_root::String;
     push!(workspaces, compiler_wrapper_path => "/opt/bin")
 
     # Generate CMake and Meson files
-    generate_toolchain_files!(platform, envs; toolchains_path=toolchains_path)
```
giordano, line 96: Side note, the constructors of the two runners have plenty of duplication :disappointed:

src/DockerRunner.jl
```diff
@@ -93,9 +93,18 @@ function DockerRunner(workspace_root::String;
...
+    generate_toolchain_files!(platform, envs, toolchains_path)
     push!(workspaces, toolchains_path => "/opt/toolchains")
 
+    # Generate directory where to write Cargo config files
+    if :rust in only(extract_kwargs(kwargs, (:compilers,))).second
+        cargo_dir = mktempdir()
+        cargo_config_file!(cargo_dir)
+        # Add to the list of mappings a subdirectory of ${CARGO_HOME}, whose content
+        # will be put in ${CARGO_HOME}.
+        push!(mappings, cargo_dir => envs["CARGO_HOME"] * "/" * randstring())
```
giordano, line 105: Do we want to preserve the `mappings` argument and create a copy of it, instead of overwriting it?  I don't think this argument is super critical though.

src/BuildToolchains.jl
```diff
@@ -253,3 +253,29 @@ function generate_toolchain_files!(platform::AbstractPlatform, envs::Dict{String
...
+            write(io, """
+            [target.$(map_rust_target(platform))]
+            linker = "$(aatriplet(platform))-gcc"
+            """)
+            if platforms_match(platform, Platform("aarch64", "linux"; libc="musl"))
+                # Work around bug for this platform:
+                # https://github.com/rust-lang/rust/issues/46651#issuecomment-433611633
+                write(io, """
+                rustflags = ["-C", "link-arg=-lgcc"]
+                """)
```
giordano, line 277: Memo to self: check whether this is still needed after having upgraded Rust to 1.56.1

src/Runner.jl
```diff
@@ -986,7 +988,7 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
         "CARGO_HOME" => "/opt/$(host_target)",
         "RUSTUP_HOME" => "/opt/$(host_target)",
         # TODO: we'll need a way to parameterize this toolchain number
-        "RUSTUP_TOOLCHAIN" => "1.43.0-$(map_rust_target(host_platform))",
+        "RUSTUP_TOOLCHAIN" => "1.56.1-$(map_rust_target(host_platform))",
```
giordano, line 991: Memo to self: the Rust and Go variables shouldn't be set if their compilers aren't requested.

giordano:
Alright, tests in https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1121 are looking great, I'm going ahead and merging this.  We may finally have a working Rust toolchain

src/DockerRunner.jl
giordano: Addressed by # 191

PR #187, opened 2021-12-03
[Rootfs] Fix name of PlatformSupport artifacts

giordano:
Also `PlatformSupport` calls the armv6l artifacts `armv7l`


giordano:
Ok, we'll catch this when we make `experimental=true` the default, soon:tm: Sorry about the mess!

PR #191, opened 2021-12-05
[Runner] Define a function to set up runner constructors

giordano:
We have lots of duplication between `UserNSRunner.jl` and `DockerRunner.jl`,
this should reduce it quite a lot and make further additions to the setup less
error-prone.


giordano:
Uhm, all tests here pass, but `run_interactive` is broken (and is clearly not tested) :thinking:

PR #193, opened 2021-12-06
[Runner] Set `-mmacosx-version-min` based on the actual macOS version

giordano:
With this I get:
```console
% julia-16 -e 'using BinaryBuilderBase; BinaryBuilderBase.runshell(Platform("aarch64", "macos"))'
sandbox:${WORKSPACE} # export SUPER_VERBOSE=1
sandbox:${WORKSPACE} # cc --version
ccache /opt/x86_64-linux-musl/bin/clang -target arm64-apple-darwin20 --sysroot=/opt/aarch64-apple-darwin20/aarch64-apple-darwin20/sys-root -march=armv8-a -mtune=cortex-a57 -nostdinc++ -isystem /opt/aarch64-apple-darwin20/aarch64-apple-darwin20/sys-root/usr/include/c++/v1 -Wno-unused-command-line-argument -mmacosx-version-min=11.0 --version -L/opt/aarch64-apple-darwin20/aarch64-apple-darwin20/lib -fuse-ld=aarch64-apple-darwin20 -headerpad_max_install_names
clang version 12.0.0 (/home/mose/.julia/dev/BinaryBuilderBase/deps/downloads/llvm-project.git d28af7c654d8db0b68c175db5ce212d74fb5e9bc)
Target: arm64-apple-darwin20
Thread model: posix
InstalledDir: /opt/x86_64-linux-musl/bin
```
Note `-mmacosx-version-min=11.0` instead of `-mmacosx-version-min=10.10`.

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -294,6 +294,7 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
     end
 
     function macos_version(kernel_version::Integer)
+        # See https://en.wikipedia.org/wiki/Darwin_(operating_system)#Release_history
         kernel_to_macos = Dict(
             12 => "10.8",
             13 => "10.9",
@@ -304,6 +305,7 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
             18 => "10.14",
             19 => "10.15",
             20 => "11.0",
+            21 => "12.0",
         )
         return get(kernel_to_macos, kernel_version, nothing)
     end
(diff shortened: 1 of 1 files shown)


giordano:
@staticfloat bump :slightly_smiling_face:

staticfloat:
I'd love to do this, but I thought that we have the triplet version number embedded in the paths to the compilers, and everything breaks if we change this?  E.g. we construct paths like `/opt/$(aatriplet(p))` which break when `triplet(p)` is `x86_64-apple-darwin15` instead of `x86_64-apple-darwin14`.  If that's solved, then great, merge away!

giordano:
Yes, the version number is in the ABI-agnostic triplet:
https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/0d1023b92f9c6d929452ee5674efad74291820e9/src/Platforms.jl#L27
And that appears in the path, you can see it in the output above (`/opt/aarch64-apple-darwin20`), but I don't see how this has anything to do with that: the functions I'm touching here only set the macosx-version-min flag

staticfloat:
Ah, okay, I thought we might call `macos_version(p)` elsewhere, but apparently not.  LGTM!

staticfloat (approved):

PR #197, opened 2021-12-29
simplify `run_interactive`, `pipeline` can now accept an `IOBuffer`

KristofferC:
```jl
julia> io = IOBuffer();

julia> run(pipeline(`echo hello`; stdout=io))
Process(`echo hello`, ProcessExited(0))

julia> String(take!(io))
"hello\n"
```

PR #199, opened 2021-12-30
Revert "simplify `run_interactive`, `pipeline` can now accept an `IOBuffer`"

KristofferC:
Reverts JuliaPackaging/BinaryBuilderBase.jl#197

This caused some issues (and we need some more tests for this apparently).


giordano (approved):

KristofferC:
Actually, it might be enough to add `IOBuffer` to https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/385c5e6c309371cbf9615fd9afd3902d666d8add/src/UserNSRunner.jl#L185.

PR #202, opened 2021-12-31
[Rootfs] Promote experimental platforms to non-experimental

giordano:
Fix https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1135


giordano:
After some discussion on Slack we agreed that it was a better idea to promote the experimental platforms to non-experimental.

Note: the API doesn't change, but now `supported_platforms()` will return more platforms than before.  Users may have to manually filter out aarch64-apple-darwin on armv6l platforms, if needed.

PR #204, opened 2022-01-09
[CI] Disable the Codecov commit statuses

DilumAluthge:


giordano:
I don't want to disable completely the check but have some tolerance to allow small decreases, like in BinaryBuilder https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/6fb9c122b7c231a124715c166a0cb80088a839ac/codecov.yml

DilumAluthge:
Ah, yeah in that case this is not the right PR.

PR #205, opened 2022-01-14
[Platforms] Fix feature flag for aarch64 microarchitectures

giordano:
https://github.com/JuliaPackaging/Yggdrasil/pull/4249 tried to use multiple
microarchitectures for the first time, but it looks like the aarch64 microarchitectures are
a bit of a mess: some flags are wrong and GCC lower bounds might also be wrong: for example
https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/ca372240dd3dd6d20d09f8571e2e3945de755dfd/src/Platforms.jl#L132-L134
doesn't match
https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/ca372240dd3dd6d20d09f8571e2e3945de755dfd/src/Rootfs.jl#L473-L476
but I didn't have the time to check these yet.


src/Platforms.jl
```diff
@@ -126,21 +135,13 @@ const ARCHITECTURE_FLAGS = Dict(
...
     ),
     "gcc" => Dict(
-        "aarch64" => Dict(
-            # `clang`/`gcc` disagree on `rdm(a)`
-            "armv8_1"            => ["-march=armv8-a+lse+crc+rdma", "-mtune=thunderx2t99"],
-            # Note that these targets require gcc 9+
-            "armv8_2_crypto"     => ["-march=armv8-a+lse+crc+rdma+aes+sha2", "-mtune=cortex-a76"],
-            "armv8_4_crypto_sve" => ["-march=armv8-a+les+crc+rdma+aes+sha2+fp16fml+dotprod+sve", "-mtune=cortex-a76"]
+        "aarch64" => OrderedDict(
+            "apple_m1"       => ["-march=armv8.5-a+aes+sha2+sha3+fp16fml+fp16+rcpc+dotprod", "-mtune=cortex-a76"],
```
giordano, line 139: I'm pretty sure `-mtune=cortex-a76` isn't great for this CPU, but at the moment GCC doesn't have a specific profile for M1.  I'll update this flag if I'll find a closer CPU profile.

PR #206, opened 2022-01-17
add manage_shards function

Crghilardi:
Following on #171, this PR adds a `manage_shards()` function that opens a prompt allowing users to choose individual shards to be removed. I originally mocked this up with a top level `RadioMenu` option for choosing install or remove first, but am still working on making the install branch work. I wanted to open this up to see if this is the right place for this and any thoughts.

### DIFF
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -2,6 +2,7 @@ export supported_platforms, expand_gfortran_versions, expand_cxxstring_abis, exp
 
 using Pkg.Artifacts: load_artifacts_toml, ensure_all_artifacts_installed
 using Base.BinaryPlatforms: set_compare_strategy!, compare_version_cap
+using REPL.TerminalMenus
 
 # This is a type that encompasses a shard; it makes it easy to pass it around,
 # get its download url, extraction url, mounting url, etc...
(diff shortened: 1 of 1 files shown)


src/Rootfs.jl
```diff
@@ -988,3 +989,17 @@ function installed_shards()
...
+    manage_shards()
+Open a prompt allowing a user to selectively remove downloaded compiler shards
+"""
+function manage_shards()
+    manage_shards_menu = TerminalMenus.request("Which compiler shards should be removed?",
+                                       TerminalMenus.MultiSelectMenu(string.(installed_shards()) , pagesize = 10; charset=:ascii)
+                                       )
+
+    installed_shards()[Int.(manage_shards_menu)]
+    Pkg.remove_artifact.(BinaryBuilderBase.shard_source_artifact_hash.(not_installed[Int.(manage_shards_menu)]))
```
sloede, line 1004: I don't understand these two lines:
The first seems to just index into an array without doing anything.
The second refers to a `not_installed` variable that does not exist.

Is there some code missing here?

Crghilardi: Probably, I opened this up at first to see if there was any interest in this type of functionality, so I might have had some psuedocode floating around as I was sketching out. 

I switched back and forth a bit starting with remove or install new shards functionality, so I might have crossed the two. There is a (unexported) function somewhere the generates a full expansion of compilers, so I might have been using that and `installed_shards` to get the differences. (`not_installed`)

Feel free to do whatever on this

Crghilardi:
Closed by #303

PR #211, opened 2022-01-27
[Prefix] Always pass `io=outs` to `Pkg.add`

giordano:


jeremiahpslewis:
@giordano does this solve a specific / known issue?

giordano:
Not showing the output of `Pkg.add` when `verbose=false`

PR #214, opened 2022-01-31
Improve relocatability of sysimages with BBBase

ven-k:
Currently `artifact_toml` paths are defined using `@__DIR__`. This hardwires the paths. Any sysimage with BinaryBuilderBase thus becomes non-relocatable. Particularly, any function that downloads artifacts throws an IOerror for not finding `<original_package_location>/Project.toml`

(Similarly `get_bbb_version` is suitably modified)

### DIFF
--- a/src/BinaryBuilderBase.jl
+++ b/src/BinaryBuilderBase.jl
@@ -74,7 +74,7 @@ const _artifacts_toml = @path Pkg.Artifacts.find_artifacts_toml(@__FILE__)
 
 function get_bbb_version(dir=@__DIR__, uuid="7f725544-6523-48cd-82d1-3fa08ff4056e")
     # Get BinaryBuilder.jl's version and git sha
-    version = Pkg.activate() do
+    version = Pkg.activate(".") do
         Pkg.Types.EnvCache().project.version
     end
     try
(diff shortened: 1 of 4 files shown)


src/BinaryBuilderBase.jl
```diff
@@ -70,9 +70,19 @@ const allow_ecryptfs = Ref(false)
 const use_ccache = Ref(false)
 const bootstrap_list = Symbol[]
 
+const _artifacts_toml = @path Pkg.Artifacts.find_artifacts_toml(@__FILE__)
+
 function get_bbb_version(dir=@__DIR__, uuid="7f725544-6523-48cd-82d1-3fa08ff4056e")
     # Get BinaryBuilder.jl's version and git sha
-    version = Pkg.TOML.parsefile(joinpath(dir, "..", "Project.toml"))["version"]
+    version = Pkg.activate(".") do
```
giordano, line 77: Note that you're ignoring `dir`.  This breaks https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/11af33aadd1d502bfd2c90a7c8382e2550439580/src/BinaryBuilder.jl#L53-L54

giordano (changes requested):

PR #216, opened 2022-02-08
[Runner] Generate some fake XCode tools which are occasionally needed

giordano:
With these two simple scripts I'm able to compile for macOS `licensecheck` without [this manual hack](https://github.com/JuliaPackaging/Yggdrasil/blob/2b0395821db522fa190e749d4d7fd670a5297643/L/licensecheck/build_tarballs.jl#L17-L24), and LLVM 13.0.1 from https://github.com/JuliaPackaging/Yggdrasil/pull/4403 without having to manually create a fake `sw_vers`

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -807,12 +807,20 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
     if os(platform) == "macos"
         # `sw_vers -productVersion` is needed to make CMake correctly initialise the macOS
         # platform, ref: <https://github.com/JuliaPackaging/Yggdrasil/pull/4403>.  In the
-        # future we may add more arguments.
+        # future we may add more arguments, see
+        # <https://www.unix.com/man-page/osx/1/sw_vers/> for reference.
+        product_version = macos_version(platform)
+        product_name = VersionNumber(product_version) < v"10.12" ? "Mac OS X" : "macOS"
         sw_vers_path = joinpath(bin_path, triplet(platform), "sw_vers")
         write(sw_vers_path, """
               #!/bin/sh
-              if [[ "\${*}" == "-productVersion" ]]; then
-                  echo "$(macos_version(platform))"
+              if [[ -z "\${@}" ]]; then
+                  echo "ProductName:    $(product_name)"
+                  echo "ProductVersion: $(product_version)"
+              elif [[ "\${@}" == "-productName" ]]; then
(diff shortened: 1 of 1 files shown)


src/Runner.jl
```diff
@@ -802,6 +802,31 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
...
+              if [[ "\${*}" == "-productVersion" ]]; then
+                  echo "$(macos_version(platform))"
+              fi
+              """)
+        chmod(sw_vers_path, 0o775)
+
+        # `xcrun` is another macOS-specific tool, which is occasionally needed to run some
+        # commands, for example for the CGO linker.  Ref:
+        # <https://github.com/JuliaPackaging/Yggdrasil/pull/2962>.
+        xcrun_path = joinpath(bin_path, triplet(platform), "xcrun")
```
Keno, line 823: Since this isn't config dependent, should we just put this script in the shard?

giordano: I'm not quite sure `xcrun` can always be emulated with just `exec "${@}"`, it worked in one case but I presume it's more complicated than that?  Having it here lets us quickly update it if/when necessary, without having to rebuild the full compiler shards, for all versions.

giordano:
New `sw_vers` in action:
```
julia> using BinaryBuilderBase

julia> BinaryBuilderBase.runshell(Platform("x86_64", "macos"));
sandbox:${WORKSPACE} # sw_vers
ProductName:    Mac OS X
ProductVersion: 10.10
sandbox:${WORKSPACE} # sw_vers -productName
Mac OS X
sandbox:${WORKSPACE} # sw_vers -productVersion
10.10
sandbox:${WORKSPACE} # 
logout

julia> BinaryBuilderBase.runshell(Platform("aarch64", "macos"));
sandbox:${WORKSPACE} # sw_vers
ProductName:    macOS
ProductVersion: 11.0
sandbox:${WORKSPACE} # sw_vers -productName
macOS
sandbox:${WORKSPACE} # sw_vers -productVersion
11.0
sandbox:${WORKSPACE} # 
logout
```
I'm ignoring the BuildVersion, that should be fine for most cases.

src/Runner.jl
staticfloat: I’d also like to use JIT’ed wrappers more and baked shards less, at least until it’s trivial to create and use such shards.

staticfloat (approved):

PR #217, opened 2022-02-09
[Runner] Set `MACOSX_DEPLOYMENT_TARGET` based on the actual macOS version

giordano:
With this I can build Enzyme for aarch64-apple-darwin without manual hacks, because
```console
% julia --compile=min -e 'using BinaryBuilderBase; BinaryBuilderBase.runshell(Platform("aarch64", "macos"))'
sandbox:${WORKSPACE} # echo $MACOSX_DEPLOYMENT_TARGET
11.0
```

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -106,6 +106,37 @@ function ld_library_path(target::AbstractPlatform,
     return join(paths, ":")
 end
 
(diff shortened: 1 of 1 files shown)


vchuravy (approved):
We are learning so.many new things about macos

PR #220, opened 2022-02-13
[Prefix] Take out function to find the git commit SHA for given git tree hash

giordano:
This function could be useful in other cases, I think it's better to take it out
from `get_addable_spec`.


giordano:
For the record, this has been used already for https://github.com/JuliaPackaging/Yggdrasil/pull/4437

PR #222, opened 2022-02-16
Fix architecture of `default_host_platform` to `x86_64`

jeremiahpslewis:
For platforms which are DockerRunner supported, use `x86_64` as the `default_host_platform` architecture.

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -15,7 +15,7 @@ const AnyRedirectable = Union{Base.AbstractCmd, Base.TTY, IOStream, IOBuffer}
 
 The default host platform in the build environment.
 """
-const default_host_platform = Sys.islinux(HostPlatform()) ? Platform(arch(HostPlatform()), "linux"; libc="musl", cxxstring_abi="cxx11") : Platform("x86_64", "linux"; libc="musl", cxxstring_abi="cxx11")
+const default_host_platform = Platform("x86_64", "linux"; libc="musl", cxxstring_abi="cxx11")
 
 function nbits(p::AbstractPlatform)
     if arch(p) in ("i686", "armv6l", "armv7l")
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -8,7 +8,8 @@ const AnyRedirectable = Union{Base.AbstractCmd, Base.TTY, IOStream, IOBuffer}
 # Host platform _must_ match the C++ string ABI of the binaries we get from the
 # repositories.  Note: when preferred_gcc_version=v"4" we can't really build for
 # that C++ string ABI :-(
-# Non-Linux will end up using DockerRunner and we need (and can use) x86 here.    
+# Non-Linux platforms *should* end up using DockerRunner
+# so we can safely assume x86_64 support
 """
     default_host_platform
 
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -8,8 +8,6 @@ const AnyRedirectable = Union{Base.AbstractCmd, Base.TTY, IOStream, IOBuffer}
 # Host platform _must_ match the C++ string ABI of the binaries we get from the
 # repositories.  Note: when preferred_gcc_version=v"4" we can't really build for
 # that C++ string ABI :-(
-# Non-Linux platforms *should* end up using DockerRunner
-# so we can safely assume x86_64 support
 """
     default_host_platform
 
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -8,12 +8,13 @@ const AnyRedirectable = Union{Base.AbstractCmd, Base.TTY, IOStream, IOBuffer}
 # Host platform _must_ match the C++ string ABI of the binaries we get from the
 # repositories.  Note: when preferred_gcc_version=v"4" we can't really build for
 # that C++ string ABI :-(
+# Non-Linux will end up using DockerRunner and we need (and can use) x86 here.    
 """
     default_host_platform
 
 The default host platform in the build environment.
 """
-const default_host_platform = Platform(arch(HostPlatform()), "linux"; libc="musl", cxxstring_abi="cxx11")
+const default_host_platform = Sys.islinux(HostPlatform()) ? Platform(arch(HostPlatform()), "linux"; libc="musl", cxxstring_abi="cxx11") : Platform("x86_64", "linux"; libc="musl", cxxstring_abi="cxx11")
 
 function nbits(p::AbstractPlatform)
     if arch(p) in ("i686", "armv6l", "armv7l")
--- a/test/dependencies.jl
+++ b/test/dependencies.jl
@@ -122,11 +122,11 @@ end
             dependencies = [
                 Dependency("Zlib_jll")
             ]
-            platform = HostPlatform()
+            platform = HostPlatform(parse(Platform, join(["x86_64", split(Base.BinaryPlatforms.host_triplet(), "-")[2:end]...], "-")))
             ap = @test_logs setup_dependencies(prefix, getpkg.(dependencies), platform)
             @test "libz." * platform_dlext(platform) in readdir(last(libdirs(Prefix(destdir(dir, platform)))))
             @test "zlib.h" in readdir(joinpath(destdir(dir, platform), "include"))
-            @test readdir(joinpath(destdir(dir, platform), "logs")) == ["Zlib.log.gz"]
+            @test "Zlib.log.gz" in readdir(joinpath(destdir(dir, platform), "logs"))
 
             # Make sure the directories are emptied by `cleanup_dependencies`
             @test_nowarn cleanup_dependencies(prefix, ap, platform)
--- a/test/dependencies.jl
+++ b/test/dependencies.jl
@@ -122,11 +122,17 @@ end
             dependencies = [
                 Dependency("Zlib_jll")
             ]
-            platform = HostPlatform(parse(Platform, join(["x86_64", split(Base.BinaryPlatforms.host_triplet(), "-")[2:end]...], "-")))
+            platform = HostPlatform()
             ap = @test_logs setup_dependencies(prefix, getpkg.(dependencies), platform)
             @test "libz." * platform_dlext(platform) in readdir(last(libdirs(Prefix(destdir(dir, platform)))))
             @test "zlib.h" in readdir(joinpath(destdir(dir, platform), "include"))
-            @test "Zlib.log.gz" in readdir(joinpath(destdir(dir, platform), "logs"))
+
+            if os(platform) == "macos"
+                zlib_log_files = ["Zlib.log.gz", "fix_identity_mismatch_libz.1.2.11.dylib.log.gz", "ldid_libz.1.2.11.dylib.log.gz"]
+            else
+                zlib_log_files = ["Zlib.log.gz"]
+            end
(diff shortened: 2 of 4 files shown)


src/Runner.jl
```diff
@@ -8,12 +8,14 @@ const AnyRedirectable = Union{Base.AbstractCmd, Base.TTY, IOStream, IOBuffer}
 # Host platform _must_ match the C++ string ABI of the binaries we get from the
 # repositories.  Note: when preferred_gcc_version=v"4" we can't really build for
 # that C++ string ABI :-(
+# Non-Linux platforms *should* end up using DockerRunner
+# so we can safely assume x86_64 support
```
giordano, line 12: This comment doesn't make much sense.  The ability to run a x86_64 doesn't depend on using Docker.
```suggestion
```

src/Runner.jl
```diff
@@ -8,12 +8,14 @@ const AnyRedirectable = Union{Base.AbstractCmd, Base.TTY, IOStream, IOBuffer}
...
 """
     default_host_platform
 
 The default host platform in the build environment.
 """
-const default_host_platform = Platform(arch(HostPlatform()), "linux"; libc="musl", cxxstring_abi="cxx11")
+const default_host_platform = Sys.islinux(HostPlatform()) ? Platform(arch(HostPlatform()), "linux"; libc="musl", cxxstring_abi="cxx11") : Platform("x86_64", "linux"; libc="musl", cxxstring_abi="cxx11")
```
giordano, line 18: I'd simply do
```suggestion
const default_host_platform = Platform("x86_64", "linux"; libc="musl", cxxstring_abi="cxx11")
```
because for the foreseeable future we aren't going to change it anyway.  _If_ and when we'll have an extra host we'll figure out a mechanism to choose it.

giordano (changes requested):
Did you actually try this and see if anything works on the M1?  I gather you didn't run the tests because I'm sure they'll fail at https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/1926acf52c1ac5ffb17f0320a3ad010198c2d2aa/test/runners.jl#L104

jeremiahpslewis:
@giordano I can confirm that this works on the m1 building actual binaries. Tried on a few, runs to completion without errors and product is created.

giordano:
But you still haven't run the tests

test/dependencies.jl
```diff
@@ -122,11 +122,11 @@ end
             dependencies = [
                 Dependency("Zlib_jll")
             ]
-            platform = HostPlatform()
+            platform = HostPlatform(parse(Platform, join(["x86_64", split(Base.BinaryPlatforms.host_triplet(), "-")[2:end]...], "-")))
```
giordano, line 125: What does mean?

test/dependencies.jl
```diff
@@ -122,11 +122,11 @@ end
...
             ap = @test_logs setup_dependencies(prefix, getpkg.(dependencies), platform)
             @test "libz." * platform_dlext(platform) in readdir(last(libdirs(Prefix(destdir(dir, platform)))))
             @test "zlib.h" in readdir(joinpath(destdir(dir, platform), "include"))
-            @test readdir(joinpath(destdir(dir, platform), "logs")) == ["Zlib.log.gz"]
+            @test "Zlib.log.gz" in readdir(joinpath(destdir(dir, platform), "logs"))
```
giordano, line 129: Why?  This isn't equivalent and the test that was before was meant to test that in that directory there is _only_ that file.

test/runners.jl
```diff
@@ -30,15 +30,17 @@ end
...
-        isecfs = (false, "/proc/")
-        @test_logs (:info, "Checking to see if /proc/ is encrypted...") @test BinaryBuilderBase.is_ecryptfs("/proc"; verbose=true) == isecfs
-        @test_logs (:info, "Checking to see if /proc/ is encrypted...") @test BinaryBuilderBase.is_ecryptfs("/proc/"; verbose=true) == isecfs
-        @test_logs (:info, "Checking to see if /proc/not_a_file is encrypted...") @test BinaryBuilderBase.is_ecryptfs("/proc/not_a_file"; verbose=true) == isecfs
-    else
-        @test_logs (:info, "Checking to see if /proc/ is encrypted...") @test BinaryBuilderBase.is_ecryptfs("/proc"; verbose=true) == (false, "/proc")
-        @test_logs (:info, "Checking to see if /proc/ is encrypted...") @test BinaryBuilderBase.is_ecryptfs("/proc/"; verbose=true) == (false, "/proc/")
-        @test_logs (:info, "Checking to see if /proc/not_a_file is encrypted...") @test BinaryBuilderBase.is_ecryptfs("/proc/not_a_file"; verbose=true) == (false, "/proc/not_a_file")
+    if Sys.islinux(HostPlatform())
+        if isdir("/proc")
```
giordano, line 34: Wouldn't
```suggestion
    if Sys.islinux() && isdir("/proc")
```
work?

test/runners.jl
```diff
@@ -101,7 +103,8 @@ end
     @testset "Compilation - C++ string ABI" begin
         mktempdir() do dir
             # Host is x86_64-linux-musl-cxx11 and target is x86_64-linux-musl-cxx03
-            ur = preferred_runner()(dir; platform=Platform(arch(HostPlatform()), "linux"; libc="musl", cxxstring_abi="cxx03"), preferred_gcc_version=v"5")
+            platform = HostPlatform(parse(Platform, join(["x86_64", split(Base.BinaryPlatforms.host_triplet(), "-")[2:end]...], "-")))
+            ur = preferred_runner()(dir; platform=Platform(arch(platform), "linux"; libc="musl", cxxstring_abi="cxx03"), preferred_gcc_version=v"5")
```
giordano, line 107: Same question as above.

giordano (changes requested):

test/runners.jl
```diff
@@ -101,7 +103,8 @@ end
     @testset "Compilation - C++ string ABI" begin
         mktempdir() do dir
             # Host is x86_64-linux-musl-cxx11 and target is x86_64-linux-musl-cxx03
-            ur = preferred_runner()(dir; platform=Platform(arch(HostPlatform()), "linux"; libc="musl", cxxstring_abi="cxx03"), preferred_gcc_version=v"5")
+            platform = HostPlatform(parse(Platform, join(["x86_64", split(Base.BinaryPlatforms.host_triplet(), "-")[2:end]...], "-")))
```
jeremiahpslewis, line 106:
```suggestion
```

test/runners.jl
jeremiahpslewis, line 107:
```suggestion
            ur = preferred_runner()(dir; platform=Platform("x86_64", "linux"; libc="musl", cxxstring_abi="cxx03"), preferred_gcc_version=v"5")
```

test/runners.jl
jeremiahpslewis: but then we hit the second half of the if statement which also fails (but is a usernsrunner check afaict)

test/dependencies.jl
jeremiahpslewis: use exactly what hostplatform gives you, but replace the arch with x86

jeremiahpslewis: it's hideous, admittedly

jeremiahpslewis:
Ok, all tests pass on M1. Cool! ;)

test/runners.jl
jeremiahpslewis: this is now fixed :)

test/dependencies.jl
jeremiahpslewis: I'll need to dig deeper tomorrow then, not sure why other files are popping up:

```Julia
Evaluated: ["Zlib.log.gz", "fix_identity_mismatch_libz.1.2.11.dylib.log.gz", "ldid_libz.1.2.11.dylib.log.gz"] == ["Zlib.log.gz"]
```

test/dependencies.jl
giordano: I think I'm missing what's wrong with keeping `HostPlatform`

test/runners.jl
giordano: It doesn't make any sense to build in a convoluted way a platform whose architecture is always `x86_64` and then at the line below pass only its architecture, forgetting all the rest.  Just use `x86_64` for the architecture of the runner.
```suggestion
            ur = preferred_runner()(dir; platform=Platform("x86_64", "linux"; libc="musl", cxxstring_abi="cxx03"), preferred_gcc_version=v"5")
```
You just need to undo https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/178/files#diff-be899a28c772a23287ab65459ca9fd65569793ca855fcb68fa63fdf0c78592d5R83

test/dependencies.jl
giordano: Bump

test/runners.jl
```diff
@@ -30,18 +30,18 @@ end
...
-        @test_logs (:info, "Checking to see if /proc/ is encrypted...") @test BinaryBuilderBase.is_ecryptfs("/proc/"; verbose=true) == isecfs
-        @test_logs (:info, "Checking to see if /proc/not_a_file is encrypted...") @test BinaryBuilderBase.is_ecryptfs("/proc/not_a_file"; verbose=true) == isecfs
-    else
-        @test_logs (:info, "Checking to see if /proc/ is encrypted...") @test BinaryBuilderBase.is_ecryptfs("/proc"; verbose=true) == (false, "/proc")
-        @test_logs (:info, "Checking to see if /proc/ is encrypted...") @test BinaryBuilderBase.is_ecryptfs("/proc/"; verbose=true) == (false, "/proc/")
-        @test_logs (:info, "Checking to see if /proc/not_a_file is encrypted...") @test BinaryBuilderBase.is_ecryptfs("/proc/not_a_file"; verbose=true) == (false, "/proc/not_a_file")
-    end
-
-    if isa(preferred_runner(), BinaryBuilderBase.DockerRunner)
+    if isa(preferred_runner(), BinaryBuilderBase.UserNSRunner)
```
giordano, line 33: I don't think this is correct.  These tests can run also if the _preferred_ runner isn't UserNS.  In fact, in CI we have the Docker job which runs these tests just fine.

test/dependencies.jl
jeremiahpslewis: OK, so it looks like the Mac builds just have more log files in the zlib binary.

test/dependencies.jl
jeremiahpslewis: Yup, you're right.

jeremiahpslewis:
Tests pass on Mac (aarch64 & x86). Manifest had to be updated so that the tests would run on aarch64

giordano (approved):
Ok, this looks better now, thanks.

Did you see that in the end you only had to modify this package and nothing in Yggdrasil? I knew the only thing to fix was the default host's architecture (which I had changed in #178 but it turned out we aren't going by that route anymore) but I never got around adjusting the tests.

> Manifest had to be updated so that the tests would run on aarch64

If you just dev the package in an environment you don't have to instantiate/update this manifest. The manifest is only used in CI (and I'd argue it's even useless to check it in here), anyway it's OK to update it.

jeremiahpslewis:
Thank you for your help diagnosing the problem, the Ygg part was a red herring.

PR #226, opened 2022-03-01
Add `-isystem ${includedir}` for clang on FreeBSD, and GCC on musl 

fingolfin:
Currently several JLLs need to manually add `-I${includedir}` to their
CPPFLAGS in order to build correctly on FreeBSD. Fix this by adjusting
the clang wrapper.

See https://github.com/JuliaPackaging/Yggdrasil/issues/3949

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -331,6 +331,23 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
         return "-mmacosx-version-min=$(macos_version(p))"
     end
 
+    function add_system_includedir(flags::Vector{String})
+        # GCC 4.8.5 for musl and clang on FreeBSD do not have `SYSROOT/usr/local` (which
+        # in our build environments is a symlink to $includedir = "$(prefix)/include")
+        # in the default list of include search directories. As a result, quite some
+        # builders that work fine everywhere else need to add something like
+        # `-I${includedir}` to their CPPFLAGS.
+        #
+        # To remove this annoyance, we simply add it here. We use `-isystem` instead
+        # of `-I` to mimic how this dir is treated on other operating systems. It
+        # is still not a perfect match, as the order of the default search directories
+        # differs, but this should not matter in practice.
+        #
+        # See also https://github.com/JuliaPackaging/Yggdrasil/issues/3949 and
+        # https://github.com/JuliaPackaging/Yggdrasil/pull/3969 for further details.
+        append!(flags, ["-isystem", raw"${includedir}"])
+    end
+
     function clang_compile_flags!(p::AbstractPlatform, flags::Vector{String} = String[])
         if lock_microarchitecture
             append!(flags, get_march_flags(arch(p), march(p), "clang"))
@@ -350,6 +367,9 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
                 "-Wno-unused-command-line-argument",
                 min_macos_version_flag(p),
             ])
+        end
+        if Sys.isfreebsd(p)
+            add_system_includedir(flags)
         end
         return flags
     end
@@ -419,6 +439,10 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
         end
         if lock_microarchitecture
             append!(flags, get_march_flags(arch(p), march(p), "gcc"))
+        end
+        gcc_version, llvm_version = select_compiler_versions(p, compilers)
+        if libc(platform) == "musl" && gcc_version == v"4.8.5"
+            add_system_includedir(flags)
         end
         return flags
     end
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -441,7 +441,7 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
             append!(flags, get_march_flags(arch(p), march(p), "gcc"))
         end
         gcc_version, llvm_version = select_compiler_versions(p, compilers)
-        if libc(platform) == "musl" && gcc_version == v"4.8.5"
+        if libc(platform) == "musl" && gcc_version in (v"4.8.5", v"5.2.0")
             add_system_includedir(flags)
         end
         return flags


src/Runner.jl
```diff
@@ -351,6 +351,9 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
                 min_macos_version_flag(p),
             ])
         end
+        if Sys.isfreebsd(p)
```
giordano, line 354: Can you please add a comment about why this is needed?  I believe you still need to do the the same for GCC on musl-based systems

fingolfin: Indeed! I've added a comment now, please proof read. Also, I had to replace `$(prefix)` by `/workspace/destdir` -- if you know a better way to avoid duplicating this here, please let me know.

I don't think we need this for GCC on musl-based systems? As far as I recall, only GCC 4.8.5 on musl had the issue, but I fixed that in https://github.com/JuliaPackaging/Yggdrasil/pull/3969

Anyway: it'll be trivial to add, so perhaps we can first test this version on Yggdrasil, and then add it if we determine it is needed after all?

src/Runner.jl
```diff
@@ -351,6 +351,22 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
...
+            # `-I${includedir}` to their CPPFLAGS when building for FreeBSD.
+            #
+            # To remove the need for this, we simply add it here. We use `-isystem`
+            # instead of `-I` to mimic what is done in clang builds for other operating
+            # systems. It is still not a perfect match, as the order of the default
+            # search directories differs, but this should not matter in practice.
+            #
+            # See also https://github.com/JuliaPackaging/Yggdrasil/issues/3949 and
+            # https://github.com/JuliaPackaging/Yggdrasil/pull/3969 for further details.
+            append!(flags, ["-isystem", "/workspace/destdir/include"])
```
fingolfin, line 368: Actually, considering what I write above, perhaps the simple way to avoid hardcoding the value of $prefix here would be to do this (and thus really mirroring what is the case elsewhere):
```suggestion
            append!(flags, ["-isystem", "/usr/local/include"])
```
Thoughts?

giordano:
```suggestion
            append!(flags, ["-isystem", raw"${includedir}"])
```
?

src/Runner.jl
giordano: > but I fixed that in https://github.com/JuliaPackaging/Yggdrasil/pull/3969

I don't think we rebuilt GCC after that, did we?

src/Runner.jl
fingolfin: Nice!

src/Runner.jl
fingolfin: Oh, somehow I thought it was. My bad then. In any case, I just opened https://github.com/JuliaPackaging/Yggdrasil/pull/4523 as I think it's best to establish a baseline now, then use that to test this PR later on.

fingolfin: Interestingly, that PR worked on FreeBSD... But it indeed failed with musl. I've now added a few more packages to that test, and also packages that do `LDFLAGS="-L${libdir}"` -- perhaps we need something similar for that, too?

fingolfin: I keep adding to that test PR, removing `-I${includedir}` statements, and so far only musl builds were broken, no FreeBSD 😂 but I know that it recently was an issue for one of the GAP_pkg_*_jlls, so it definitely still is there (I'll try to repro it now with one of those).

There are also various mingw failures, but I think they are unrelated?

fingolfin: Updated to check for musl

src/Runner.jl
```diff
@@ -420,6 +440,9 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
         if lock_microarchitecture
             append!(flags, get_march_flags(arch(p), march(p), "gcc"))
         end
+        if libc(platform) == "musl"
```
giordano, line 443: This should be needed only for GCC 4.8.5, right?  You can check the version number like in https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/c3a40125ecdb2869c8d0355389f88122fd89d1e8/src/Runner.jl#L429-L430

giordano (commented):
Last comment, then it should be good to go

src/Runner.jl
fingolfin: Added `gcc_version == v"4.8.5"` check

fingolfin: @giordano OK now?

giordano: `gcc_version` is undefined there, or am I missing something?

fingolfin: Argh, indeed. I've now added
```julia
        gcc_version, llvm_version = select_compiler_versions(p, compilers)
```

giordano: Mumble mumble mumble... this also means this line is untested, which is very weird to be honest :confused:

src/Runner.jl
```diff
@@ -420,6 +440,10 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
         if lock_microarchitecture
             append!(flags, get_march_flags(arch(p), march(p), "gcc"))
         end
+        gcc_version, llvm_version = select_compiler_versions(p, compilers)
+        if libc(platform) == "musl" && gcc_version == v"4.8.5"
```
giordano, line 444: I went through the output of `cc -E -v - < /dev/null` of all versions of GCC from 4 to 11, it looks like we need this also for GCC 5:
```suggestion
        if libc(platform) == "musl" && gcc_version in (v"4.8.5", v"5.2.0")
```

src/Runner.jl
giordano: Nevermind, it didn't error because there is a function called `gcc_version`: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/c3a40125ecdb2869c8d0355389f88122fd89d1e8/src/Rootfs.jl#L433 :grimacing:

fingolfin: good that you paid close attention then 😅

src/Runner.jl
```diff
@@ -331,6 +331,23 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
...
+        # `-I${includedir}` to their CPPFLAGS.
+        #
+        # To remove this annoyance, we simply add it here. We use `-isystem` instead
+        # of `-I` to mimic how this dir is treated on other operating systems. It
+        # is still not a perfect match, as the order of the default search directories
+        # differs, but this should not matter in practice.
+        #
+        # See also https://github.com/JuliaPackaging/Yggdrasil/issues/3949 and
+        # https://github.com/JuliaPackaging/Yggdrasil/pull/3969 for further details.
+        append!(flags, ["-isystem", raw"${includedir}"])
```
giordano, line 348: I was looking at https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/9c15402d496a737a5d34b01409044a52c642f9be/src/BuildToolchains.jl#L114 and I realised this isn't right. We should use host_includedir if this is the host compiler (and I'm not sure we have an easy way to check that at the moment)

PR #233, opened 2022-03-23
[Products] Don't dlopen libraries built for an incompatible ISA

giordano:
Exactly same thing as https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1194 (which will eventually be based on this PR).  I still need to add tests, once I figure out how `platforms_match` is supposed to compare platforms.


src/utils.jl
```diff
@@ -60,3 +63,45 @@ end
...
+        # ISA `b` is compatible with ISA `a` only if it's a subset of `a`
+        return b_isa ≤ a_isa
+    else
+        # ISA `a` is compatible with ISA `b` only if it's a subset of `b`
+        return a_isa ≤ b_isa
+    end
+end
+
+function augment_microarchitecture!(platform::Platform)
+    haskey(platform, "march") && return platform
```
vchuravy, line 99: You might still need to call `set_compare_strategy` since I believe we can get here by just parting the triple

giordano: To be clear, you're suggesting to apply `set_compare_strategy!(platform, "march", march_comparison_strategy)` before returning `platform`?

vchuravy: Yes.

test/compat.jl
```diff
@@ -71,3 +71,44 @@ using BinaryBuilderBase: download_verify, list_tarball_files
         end
     end
 end
+
+using BinaryBuilderBase: march_comparison_strategy
+using Base.BinaryPlatforms: Platform, platforms_match, set_compare_strategy!
+@testset "Microarchitecture augmentation" begin
+    linux_x86_64 = Platform("x86_64", "linux")
+    linux_avx = Platform("x86_64", "linux"; march="avx")
```
vchuravy, line 79: I am not sure I would call this `march`. `llc` has `mcpu` which has names like `opteron` or `x86-64-v3` and `mattr` which allows you to set feature sets. 

`avx` is the name of a feature set. I think following hwcaps/linux feature level and only support:

```
x86-64-v1
x86-64-v2
x86-64-v3
x86-64-v4
```

Might be the best.
x-ref: https://github.com/JuliaLang/julia/issues/42073

The logic for how to map `x86-64-v3` to different compilers is defined here https://github.com/archspec/archspec-json/blob/020e418c9fe991a8155e53245fde4e4bdef08382/cpu/microarchitectures.json#L135-L191

giordano: He, those names are in `Base`: https://github.com/JuliaLang/julia/blob/c50361144ce783cecf65294e09d6e7cfdc73eb3f/base/binaryplatforms.jl#L598 changing them is quite a nightmare

giordano: ...Unless you're going to suggest to move the dictionary to https://github.com/JuliaPackaging/Yggdrasil/blob/master/platforms/microarchitectures.jl

vchuravy: :/ How are we going to deal with adding new names? E.g. `power9`? I would prefer for these not to be dependent on Julia releases. So yes moving them there or MicroArchitectures.jl (better than JLLWrappers?) would be nice.

giordano: I think we still depend on julia for features detection (and for powerpc that's completely missing).

giordano: To be clear, what name do you suggest instead of `march`?  I think we initially thought about `microarchitecture` but that was a bit mouthful.

I'm also mostly ok with revising the levels we have (especially if we move them to https://github.com/JuliaPackaging/Yggdrasil/blob/master/platforms/microarchitectures.jl so we don't need to worry about compatibility), but we'll also need to revise all the compiler flags and versions in https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/55c795dbca6c74b076636bdcfdccbfeb34063eb7/src/Platforms.jl#L90-L147 which I wasn't exactly looking forward to :grin:  Side note, I'm moderately sure we settled on the sets we now have in Julia Base before the glibc x86-64 levels were widely advertised :wink:

vchuravy: I think `march` is fine, I am more opposed to calling `avx` a micro-architecture since it is not, but a micro-architecture feature.

Yeah which is why I pointed out archspec, we could parse that JSON instead of maintaining the same thing in BB.

PR #234, opened 2022-04-05
allow for uuid to be passed to get_Addable_spec

vchuravy:


giordano (approved):
This is ok with me, but I think this is working correctly in https://github.com/JuliaPackaging/Yggdrasil/pull/4711
```julia
julia> ctx = Pkg.Types.Context();

julia> name = "SuiteSparse_jll";

julia> first(Pkg.Types.registry_resolve!(ctx.registries, Pkg.Types.PackageSpec(;name))).uuid
UUID("bea87d4a-7f5b-5778-9afe-8cc45184846c")
```

src/Prefix.jl
```diff
@@ -462,9 +462,8 @@ Given a JLL name and registered version, return a `PackageSpec` that, when passe
     ]
 """
 function get_addable_spec(name::AbstractString, version::VersionNumber;
+                          uuid = first(Pkg.Types.registry_resolve!(ctx.registries, Pkg.Types.PackageSpec(;name))).uuid,
                           ctx = Pkg.Types.Context(), verbose::Bool = false)
```
giordano, line 466:
```suggestion
                          ctx = Pkg.Types.Context(),
                          uuid = first(Pkg.Types.registry_resolve!(ctx.registries, Pkg.Types.PackageSpec(;name))).uuid,
                          verbose::Bool = false)
```

giordano:
I don't think this was really necessary after all.  The real problem was that `get_addable_spec` wasn't updating the registry before looking for the new version, which is resolved by #252, and now with https://github.com/JuliaPackaging/Yggdrasil/pull/5078 we also use the eager registry in Yggdrasil.  This should hopefully solve all issues with `get_addable_spec` failing to resolve the dependencies.

PR #237, opened 2022-04-15
Bump Rootfs to include new Musl build

staticfloat:
Includes https://github.com/JuliaPackaging/Yggdrasil/pull/4762 and https://github.com/JuliaPackaging/Yggdrasil/pull/4753/commits/c7bd98f77b6f14b9ac42c78a99d4ef659b83bbf3

### DIFF
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -544,7 +544,8 @@ consists of four shards, but that may not always be the case.
 """
 function choose_shards(p::AbstractPlatform;
             compilers::Vector{Symbol} = [:c],
-            rootfs_build::VersionNumber=v"2022.3.5",
+            # We always just use the latest Rootfs embedded within our Artifacts.toml
+            rootfs_build::VersionNumber=last(BinaryBuilderBase.get_available_builds("Rootfs")),
             ps_build::VersionNumber=v"2021.08.10",
             GCC_builds::Vector{GCCBuild}=available_gcc_builds,
             LLVM_builds::Vector{LLVMBuild}=available_llvm_builds,
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -2858,27 +2858,27 @@ os = "linux"
     sha256 = "5cc8c42875360e9ff28e49135556363bdb43a837d660fbd86fafe64529d2abcd"
     url = "https://github.com/JuliaPackaging/Yggdrasil/releases/download/PlatformSupport-v2021.8.10/PlatformSupport-x86_64-w64-mingw32.v2021.8.10.x86_64-linux-musl.unpacked.tar.gz"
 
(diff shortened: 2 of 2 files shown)


giordano:
Did you update https://github.com/JuliaPackaging/Yggdrasil/blob/f99f62061e9add8e288fca35673f267318d470df/0_RootFS/Rootfs/bundled/libs/libc/download_libcs.sh#L15?

src/Rootfs.jl
```diff
@@ -544,7 +544,7 @@ consists of four shards, but that may not always be the case.
 """
 function choose_shards(p::AbstractPlatform;
             compilers::Vector{Symbol} = [:c],
-            rootfs_build::VersionNumber=v"2022.3.5",
+            rootfs_build::VersionNumber=v"2022.4.15",
```
giordano, line 547: We should be able to use the new https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/67ad9e77d2c31e51681b7b9b31c480b056bec4ab/src/Rootfs.jl#L387-L388 to automatically get the version number

giordano (approved):
The changes look good to me and I confirm the sonames of the musl libcs are indeed fixed:
```console
sandbox:${WORKSPACE} # readelf -d /lib/ld-musl-* | grep SONAME
 0x0000000e (SONAME)                     Library soname: [libc.musl-i386.so.1]
 0x000000000000000e (SONAME)             Library soname: [libc.musl-x86_64.so.1]
```
On `master` it was the bugged as you described in the Yggdrasil PR:
```console
sandbox:${WORKSPACE} # readelf -d /lib/ld-musl-* | grep SONAME
 0x0000000e (SONAME)                     Library soname: [libc.musl-.so.1]
 0x000000000000000e (SONAME)             Library soname: [libc.musl-.so.1]
```

However I don't understand the failure on nightly.  I _think_ it should be unrelated to this PR, but it started only with this PR.  I'll try to have a look.

giordano:
Ok, #238 is failing on nightly in the same way, so this PR didn't break anything.  Now we have to find what broke installation of the dependencies.... Last successful CI run is https://github.com/JuliaPackaging/BinaryBuilderBase.jl/runs/6030958833?check_suite_focus=true, which used
```
Julia Version 1.9.0-DEV.382
Commit fbec395fd0 (2022-04-14 08:19 UTC)
```
First failing job is https://github.com/JuliaPackaging/BinaryBuilderBase.jl/runs/6035628989?check_suite_focus=true, which used
```
Julia Version 1.9.0-DEV.395
Commit 4f722541556 (2022-04-15 05:47 UTC)
```
Diff: https://github.com/JuliaLang/julia/compare/JuliaLang:fbec395...JuliaLang:4f72254, which includes https://github.com/JuliaLang/julia/pull/44985

giordano:
Should be fixed by #239

PR #247, opened 2022-06-05
[Runner] Manually set some search paths for Clang on Windows

giordano:
This isn't a nice solution (to say the least), but I don't understand why
`--gcc-toolchain` above isn't sufficient.  We should definitely not need to set
all these search paths manually, it's fragile at best, but at least it works for
the time being.


giordano:
Superseded by #324

PR #249, opened 2022-06-10
[Dependencies] Define new `RuntimeDependency`

giordano:
I quickly prepared this.  Still needs test and I have to see whether anything else is needed in BinaryBuilder (hopefully not).

CC: @vchuravy


giordano:
Ok, this should be ready on this side.  It turns out I do need to do a change in BinaryBuilder, but only a small one (I simply forgot to use the `is_build_dependency` trait in one place, this was unnecessary so far because all current dependency types are build dependencies).  Will do in a follow-up PR.

vchuravy (approved):

PR #250, opened 2022-06-16
[Prefix] Work around the fact `Pkg.add` modifies dependencies in-place

giordano:
CC: @Alexander-Barth.  Fix https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1210.


Alexander-Barth:
Thanks a lot!

PR #258, opened 2022-08-24
Also make sanitizer use clang on cmake

gbaraldi:
Cmake was trying to use gcc with sanitizers enabled

### DIFF
--- a/src/BuildToolchains.jl
+++ b/src/BuildToolchains.jl
@@ -234,7 +234,7 @@ function generate_toolchain_files!(platform::AbstractPlatform, envs::Dict{String
         symlink_if_exists(target, link) = ispath(joinpath(dir, target)) && symlink(target, link)
 
         # On FreeBSD and MacOS we actually want to default to clang, otherwise gcc
-        if Sys.isbsd(p)
+        if Sys.isbsd(p) || sanitize(p) in ("memory", "address")
             symlink_if_exists("host_$(aatriplet(p))_clang.cmake", joinpath(dir, "host_$(aatriplet(p)).cmake"))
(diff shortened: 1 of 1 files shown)


src/BuildToolchains.jl
```diff
@@ -234,7 +234,7 @@ function generate_toolchain_files!(platform::AbstractPlatform, envs::Dict{String
         symlink_if_exists(target, link) = ispath(joinpath(dir, target)) && symlink(target, link)
 
         # On FreeBSD and MacOS we actually want to default to clang, otherwise gcc
-        if Sys.isbsd(p)
+        if Sys.isbsd(p) || sanitize(p) in ("memory", "address")
```
giordano, line 237: Actually: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/ebbc1d8943ebefb0a9ac2d74a445f63468c5eecb/src/Runner.jl#L558 (from #255)

gbaraldi: Oh, oops

src/BuildToolchains.jl
```diff
@@ -234,7 +234,7 @@ function generate_toolchain_files!(platform::AbstractPlatform, envs::Dict{String
         symlink_if_exists(target, link) = ispath(joinpath(dir, target)) && symlink(target, link)
 
         # On FreeBSD and MacOS we actually want to default to clang, otherwise gcc
-        if Sys.isbsd(p)
+        if Sys.isbsd(p) || sanitize(p) in ("memory", "memory_origins", "address")
```
Keno, line 237: Should we make a `prefer_clang(p)` helper rather than copy-pasting this everywhere?

Issue #248, opened 2022-06-05
Clang wrapper for Windows isn't working


giordano:
These are some of the issues of our Clang wrapper for Windows:
```console
% julia --compile=min -e 'using BinaryBuilderBase; BinaryBuilderBase.runshell(Platform("x86_64", "windows"))'                          
sandbox:${WORKSPACE} # echo 'int main(){return 0;}' | clang -x c -
clang-13: warning: argument unused during compilation: '--gcc-toolchain=/opt/x86_64-w64-mingw32' [-Wunused-command-line-argument]
clang-13: warning: argument unused during compilation: '-stdlib=libstdc++' [-Wunused-command-line-argument]
/opt/x86_64-w64-mingw32/bin/x86_64-w64-mingw32-ld: cannot find -lgcc
/opt/x86_64-w64-mingw32/bin/x86_64-w64-mingw32-ld: cannot find -lgcc_eh
/opt/x86_64-w64-mingw32/bin/x86_64-w64-mingw32-ld: cannot find -lgcc
/opt/x86_64-w64-mingw32/bin/x86_64-w64-mingw32-ld: cannot find -lgcc_eh
clang-13: error: linker command failed with exit code 1 (use -v to see invocation)
sandbox:${WORKSPACE} # echo -e '#include <limits>\nint main(){return 0;}' | clang -x c++ -
clang-13: warning: argument unused during compilation: '--gcc-toolchain=/opt/x86_64-w64-mingw32' [-Wunused-command-line-argument]
<stdin>:1:10: fatal error: 'limits' file not found
#include <limits>
         ^~~~~~~~
1 error generated.
```
We actually already do https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/4b21cde2105830e287271c323d909b9ee8a64c5e/src/Runner.jl#L318-L319 to make Clang find the GCC toolchain, but it doesn't seem to be effective on Windows, maybe Clang expects a different tree structure, or maybe it's just ignored (why?), given the warning about the argument being unused.

I tried to address this issue in #247, but it's a horrible ad-hoc kludge.  I opened this issue because I feel like we should be able to make this work more nicely than putting a bunch of random `-I...` and `-L...` arguments.

giordano:
[...]

PR #261, opened 2022-08-25
Move -rtlib and friends to linker flags

gbaraldi:
This is to avoid unused argument warnings like in [this build](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=21607&view=logs&jobId=5850b9b5-9a51-573f-7d03-f1fedbd8a7fb&j=5850b9b5-9a51-573f-7d03-f1fedbd8a7fb&t=34cffb34-e22b-5097-6aba-496cb946d943)

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -321,12 +321,6 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
             # Set our sysroot to the platform-specific location, dropping compiler ABI annotations
             "--sysroot=/opt/$(aatriplet(p))/$(aatriplet(p))/sys-root",
         ])
-        # For MacOS and FreeBSD, we don't set `-rtlib`, and FreeBSD is special-cased within the LLVM source tree
-        # to not allow for -gcc-toolchain, which means that we have to manually add the location of libgcc_s.  LE SIGH.
-        # We do that within `clang_linker_flags()`, so that we don't get "unused argument" warnings all over the place.
-        # https://github.com/llvm-mirror/clang/blob/f3b7928366f63b51ffc97e74f8afcff497c57e8d/lib/Driver/ToolChains/FreeBSD.cpp
-        # For everything else, we provide `-rtlib=libgcc` because clang-builtins are broken (pending Valentin-based-magic),
-        # and we also need to provide `-stdlib=libstdc++` to match Julia on these platforms.
         if !Sys.isbsd(p)
             append!(flags, [
                 # Find GCC toolchain here (for things like libgcc_s)
@@ -420,6 +414,12 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
         if Sys.isbsd(p)
             push!(flags, "-L/opt/$(aatriplet(p))/$(aatriplet(p))/lib")
         end
+        # For MacOS and FreeBSD, we don't set `-rtlib`, and FreeBSD is special-cased within the LLVM source tree
+        # to not allow for -gcc-toolchain, which means that we have to manually add the location of libgcc_s.  LE SIGH.
+        # We do that here, so that we don't get "unused argument" warnings all over the place.
+        # https://github.com/llvm-mirror/clang/blob/f3b7928366f63b51ffc97e74f8afcff497c57e8d/lib/Driver/ToolChains/FreeBSD.cpp
+        # For everything else, we provide `-rtlib=libgcc` because clang-builtins are broken (pending Valentin-based-magic),
+        # and we also need to provide `-stdlib=libstdc++` to match Julia on these platforms.
         if !Sys.isbsd(p)
             append!(flags, [
                 # Use libgcc as the C runtime library
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -331,10 +331,6 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
             append!(flags, [
                 # Find GCC toolchain here (for things like libgcc_s)
                 "--gcc-toolchain=/opt/$(aatriplet(p))"
-                # Use libgcc as the C runtime library
-                "-rtlib=libgcc"
-                # Use libstdc++ as the C++ runtime library
-                "-stdlib=libstdc++"
             ])
         end
         return flags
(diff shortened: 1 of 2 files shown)


src/Runner.jl
```diff
@@ -331,10 +331,6 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
             append!(flags, [
                 # Find GCC toolchain here (for things like libgcc_s)
                 "--gcc-toolchain=/opt/$(aatriplet(p))"
-                # Use libgcc as the C runtime library
-                "-rtlib=libgcc"
-                # Use libstdc++ as the C++ runtime library
-                "-stdlib=libstdc++"
```
giordano, line 337: The comments above reference `-rtlib` and `-stdlib` which now don't make much sense.  Should rearrange them as well.

test/runners.jl
```diff
@@ -346,3 +370,6 @@ end
         end
     end
 end
+
+
+
```
giordano, line 375: Too much space here?

test/runners.jl
```diff
@@ -123,6 +123,30 @@ end
...
+            }
+            """
+            test_script = """
+            set -e
+            echo '$(test_c)' > test.c
+            clang -Werror -c test.c
+            clang -Werror -shared test.c -o test.\${dlext}
+            """
+            cmd = `/bin/bash -c "$(test_script)"`
+            @test run(ur, cmd, iobuff; tee_stream=devnull) broken=Sys.iswindows(platform)
```
giordano, line 143: May also be good to explain that Windows is broken because of #248

gbaraldi: Yeah, it needs some polish still. I just want to pass tests first ;)

gbaraldi: Has armv7/6 ever worked on clang? Or do I just disable the test on that platform?

gbaraldi: It fails on musl only. Also windows passes this test apparently :)

giordano:
```console
% julia --compile=min -e 'using BinaryBuilderBase; BinaryBuilderBase.runshell(Platform("armv7l", "linux"; libc="musl"))'
sandbox:${WORKSPACE} # echo 'int test(){return 0;}' | clang -x c - -c -o test.o
clang-13: warning: argument unused during compilation: '-rtlib=libgcc' [-Wunused-command-line-argument]
clang-13: warning: argument unused during compilation: '-stdlib=libstdc++' [-Wunused-command-line-argument]
sandbox:${WORKSPACE} # echo 'int test(){return 0;}' | clang -x c - -shared -o test.so
/opt/arm-linux-musleabihf/bin/arm-linux-musleabihf-ld: cannot find crtbeginS.o: No such file or directory
/opt/arm-linux-musleabihf/bin/arm-linux-musleabihf-ld: cannot find -lgcc
/opt/arm-linux-musleabihf/bin/arm-linux-musleabihf-ld: cannot find -lgcc_s
/opt/arm-linux-musleabihf/bin/arm-linux-musleabihf-ld: cannot find -lgcc
/opt/arm-linux-musleabihf/bin/arm-linux-musleabihf-ld: cannot find -lgcc_s
/opt/arm-linux-musleabihf/bin/arm-linux-musleabihf-ld: cannot find crtendS.o: No such file or directory
clang-13: error: linker command failed with exit code 1 (use -v to see invocation)
sandbox:${WORKSPACE} # qfind /opt/${target}/${target} -name 'libgcc_s.*'
/opt/arm-linux-musleabihf/arm-linux-musleabihf/lib/libgcc_s.so
/opt/arm-linux-musleabihf/arm-linux-musleabihf/lib/libgcc_s.so.1
```
uhm, the linker seems subtly broken, no idea why, libraries are of course there (otherwise gcc wouldn't work either).  Mind opening another issue and mark also those platforms as broken?

test/runners.jl
```diff
@@ -123,6 +123,30 @@ end
...
+            }
+            """
+            test_script = """
+            set -e
+            echo '$(test_c)' > test.c
+            clang -Werror -c test.c
+            clang -Werror -shared test.c -o test.\${dlext}
+            """
+            cmd = `/bin/bash -c "$(test_script)"`
+            @test run(ur, cmd, iobuff; tee_stream=devnull) broken=platform in [Platform("armv7l", "linux"; libc="musl"), Platform("armv6l", "linux"; libc="musl")]
```
giordano, line 143: Define a variable `broken_platform` and use it below (where you also used the wrong syntax)?

gbaraldi: Doing development in class means brain lag

test/runners.jl
```diff
@@ -123,6 +123,31 @@ end
         end
     end
 
+    # Test that we get no warnings when compiling without linking and when building a shared lib with clang
+    @testset "Clang - $(platform)" for platform in platforms
+        mktempdir() do dir
+            is_broken = platform in [Platform("armv7l", "linux"; libc="musl"), Platform("armv6l", "linux"; libc="musl")]
```
giordano, line 129: Windows _is_ broken, and arm-musl does have "error"/"warning" in the output.  Also, can we get the links to #248 and #262 in the comments, please? :slightly_smiling_face:

gbaraldi: That's what I'm doing ;|

test/runners.jl
```diff
@@ -123,6 +123,32 @@ end
...
+            set -e
+            echo '$(test_c)' > test.c
+            clang -Werror -c test.c
+            clang -Werror -shared test.c -o test.\${dlext}
+            """
+            cmd = `/bin/bash -c "$(test_script)"`
+            @test run(ur, cmd, iobuff; tee_stream=devnull) broken=is_broken
+            seekstart(iobuff)
+            output = String(read(iobuff))
+            @test (occursin("error", output) || occursin("warning", output)) skip=is_broken
```
giordano, line 148: I'm quite confused by this test to be honest.  I don't expect to see anything in the output, I'm having a hard time to understand what this is supposed to be doing,

giordano: Right, `output` actually contains also the input string, so "error" is _always_ there.  This test as is isn't very useful.  Not sure there is anything worth checking in the output, the above `@test` may be enough for this case.

gbaraldi: Yeah, I got confused about that at first because I didn't expect that. But I guess the first test is already enough due to -Werror

giordano:
This broke `clang++` when compiling object files without linking:
```console
sandbox:${WORKSPACE}/srcdir/SuiteSparse # cat foo.cpp 
```
```c++
#include <iostream>
std::complex<double> add(std::complex<double> a, std::complex<double> b) {
    return a + b;
}
```
```console
sandbox:${WORKSPACE}/srcdir/SuiteSparse # clang++ -c foo.cpp 
In file included from foo.cpp:1:
In file included from /opt/x86_64-linux-musl/bin/../include/c++/v1/iostream:37:
In file included from /opt/x86_64-linux-musl/bin/../include/c++/v1/ios:214:
In file included from /opt/x86_64-linux-musl/bin/../include/c++/v1/__locale:47:
/opt/x86_64-linux-musl/bin/../include/c++/v1/__support/musl/xlocale.h:27:25: error: static declaration of 'strtoll_l' follows non-static declaration
static inline long long strtoll_l(const char *nptr, char **endptr, int base,
                        ^
[...]
```

`-stdlib=libstdc++` should have been kept, but only when compiling C++, not C code (hence the warning).  At least we've got a simple reproducer for the tests :slightly_smiling_face:

gbaraldi:
Is there a function that defines `clang++` flags instea of all clangs?

PR #264, opened 2022-08-27
[Prefix] Force "sanitize" tag to platform when installing deps

giordano:


src/Prefix.jl
```diff
@@ -568,6 +568,16 @@ to modify the dependent artifact files, and (c) keeping a record of what files a
 dependencies as opposed to the package being built, in the form of symlinks to a specific artifacts
 directory.
 """
+# During installation of the artifacts we may want to enforce the platform has specific
+# properties, for example a non-empty "sanitize" tag.
+function normalize_platform(p::Platform)
+    new_p = deepcopy(p)
+    new_p["sanitize"] = get(new_p.tags, "sanitize", "none")
```
Keno, line 575: I can see why this works, but wouldn't it be cleaner to just change the comparison function in the platform object here?

giordano: In order to add a strategy comparison, the platform has to have the tag in the first place:
```julia
julia> using Base.BinaryPlatforms

julia> p = HostPlatform()
Linux x86_64 {cxxstring_abi=cxx11, julia_version=1.8.0, libc=glibc, libgfortran_version=5.0.0, libstdcxx_version=3.4.29}

julia> BinaryPlatforms.set_compare_strategy!(p, "sanitize", identity)
ERROR: ArgumentError: Cannot set comparison strategy for nonexistant tag sanitize!
Stacktrace:
 [1] set_compare_strategy!(p::Platform, key::String, f::Function)
   @ Base.BinaryPlatforms ./binaryplatforms.jl:262
 [2] top-level scope
   @ REPL[21]:1
```

Keno: I see :/ that doesn't seem like a particularly well thought out design, but I guess we're stuck with it.

giordano: I'm not sure we're stuck forever with this because I don't think this is public API, so we can change it.

I think the idea was that only when the tag is added a comparison strategy can be set, to avoid conflicting strategies, if they're set by different consumers.  @staticfloat does that make sense?

giordano:
Although we may want to improve `Base.BinaryPlatforms` API in the future, for the time being we're going ahead with this change here.

jheinen:
Using `BinaryBuilderBase#master` fixed the problem - thanks!

PR #271, opened 2022-09-23
Make the platform expansions MSAN aware

gbaraldi:
How about this, so then we don't get useless builds, but still have the tags when needed

### DIFF
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -750,7 +750,7 @@ function expand_gfortran_versions(platform::AbstractPlatform)
     end
 
     if sanitize(platform) == "memory"
-        return platform #MSAN doesn't use libgfortran, it uses libflang
+        return [platform] #MSAN doesn't use libgfortran, it uses libflang
     end
 
     # If this is an platform that has limited GCC support (such as aarch64-apple-darwin),
@@ -796,7 +796,7 @@ function expand_cxxstring_abis(platform::AbstractPlatform; skip=Sys.isbsd)
     if sanitize(platform) == "memory"
         p = deepcopy(platform)
         p["cxxstring_abi"] = "cxx11" #Clang only seems to generate cxx11 abi
-        return p
+        return [p]
     end
                                                                                                                                  
     # Otherwise, generate new versions!
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -750,7 +750,7 @@ function expand_gfortran_versions(platform::AbstractPlatform)
     end
 
     if sanitize(platform) == "memory"
-        return platform #MSAN can't use libgfortran
+        return platform #MSAN doesn't use libgfortran, it uses libflang
     end
 
     # If this is an platform that has limited GCC support (such as aarch64-apple-darwin),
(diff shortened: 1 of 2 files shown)


src/Rootfs.jl
```diff
@@ -789,6 +793,12 @@ function expand_cxxstring_abis(platform::AbstractPlatform; skip=Sys.isbsd)
         return [platform]
     end
 
+    if sanitize(platform) == "memory"
+        p = deepcopy(platform)
+        p["cxxstring_abi"] = "cxx11" #Clang only seems to generate cxx11 abi
+        return p
+    end
```
giordano, line 800: Add tests

src/Rootfs.jl
```diff
@@ -749,6 +749,10 @@ function expand_gfortran_versions(platform::AbstractPlatform)
         return [platform]
     end
 
+    if sanitize(platform) == "memory"
+        return platform #MSAN can't use libgfortran
+    end
```
giordano, line 754: And for this as well.

src/Rootfs.jl
```diff
@@ -749,6 +749,10 @@ function expand_gfortran_versions(platform::AbstractPlatform)
         return [platform]
     end
 
+    if sanitize(platform) == "memory"
+        return platform #MSAN can't use libgfortran
```
giordano, line 753: I'd say "doesn't use", rather than "can't use", it just uses a different fortran runtime library

src/Rootfs.jl
```diff
@@ -789,6 +793,12 @@ function expand_cxxstring_abis(platform::AbstractPlatform; skip=Sys.isbsd)
         return [platform]
     end
 
+    if sanitize(platform) == "memory"
+        p = deepcopy(platform)
+        p["cxxstring_abi"] = "cxx11" #Clang only seems to generate cxx11 abi
+        return p
```
giordano, line 799:
```suggestion
        return [p]
```

giordano (commented):
Good that I asked to add tests :slightly_smiling_face:

src/Rootfs.jl
```diff
@@ -749,6 +749,10 @@ function expand_gfortran_versions(platform::AbstractPlatform)
         return [platform]
     end
 
+    if sanitize(platform) == "memory"
+        return platform #MSAN doesn't use libgfortran, it uses libflang
```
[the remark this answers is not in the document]
gbaraldi: That's the actual name funilly enough https://github.com/JuliaPackaging/Yggdrasil/blob/master/F/FlangClassic/FlangClassic_RTLib/build_tarballs.jl. 
At least for FlangClassic which is what we use.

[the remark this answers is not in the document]
giordano: Right, I was looking at the wrong flang.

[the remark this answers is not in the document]
gbaraldi: In fact the OpenBlas recipe is wrong right now, it's not just a build dep but a runtime one too

gbaraldi:
This made it type unstable for some reason, which led to the issue above.

PR #272, opened 2022-09-25
Enforce type stability and test for it

gbaraldi:
#271 made `expand_cxxstring_abis` type unstable, so force specialization on it.


test/rootfs.jl
```diff
@@ -27,6 +27,7 @@ using BinaryBuilderBase
     ]
     @test expand_gfortran_versions([Platform("x86_64", "linux"; sanitize="memory")]) ==
         [Platform("x86_64", "linux"; sanitize="memory")]
+    @test expand_gfortran_versions(Platform[]) == Platform[]
```
giordano, line 30: This isn't a good test:
```julia
julia> [] == Platform[]
true
```
You want to check either the `eltype` or, which is pretty much the same, `x isa Vector{Platform}`

PR #275, opened 2022-09-29
Add the concept of a top-level dependency

vchuravy:
Motivated by https://github.com/JuliaParallel/MPI.jl/issues/646

In order for invalidation to happen we need to declare the dependency on the top-level of the file instead of in one of the platform specific wrappers.

### DIFF
--- a/src/Dependencies.jl
+++ b/src/Dependencies.jl
@@ -1,7 +1,7 @@
 using UUIDs
 
 export Dependency, RuntimeDependency, BuildDependency, HostBuildDependency,
-    is_host_dependency, is_target_dependency, is_build_dependency, is_runtime_dependency,
+    is_host_dependency, is_target_dependency, is_build_dependency, is_runtime_dependency, is_top_level_dependency,
     filter_platforms
 
 
@@ -32,6 +32,7 @@ Subtypes of `AbstractDependency` should define the following traits:
 * [`is_target_dependency`](@ref)
 * [`is_build_dependency`](@ref)
 * [`is_runtime_dependency`](@ref)
+* [`is_top_level_dependency`][@ref]
 """
 abstract type AbstractDependency end
 
@@ -64,6 +65,13 @@ Return whether `dep` is a runtime dependency or not.
 is_runtime_dependency
 
 """
+    is_top_level_dependency(dep::AbstractDependency) ->
+
+Return wheter `dep` is a top-level dependency or not.
+"""
+is_top_level_dependency(dep::AbstractDependency) = false
+
+"""
     Dependency(dep::Union{PackageSpec,String}, build_version::VersionNumber;
                compat::String, platforms::Vector{<:AbstractPlatform})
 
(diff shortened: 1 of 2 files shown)


src/Dependencies.jl
```diff
@@ -92,8 +100,10 @@ struct Dependency <: AbstractDependency
     build_version::Union{VersionNumber,Nothing}
     compat::String  # semver string for use in Project.toml of the JLL
     platforms::Vector{<:AbstractPlatform}
+    top_level::Bool
```
giordano, line 103: I think this may need to be propagated in the JSON file?

giordano (commented):
Can you please elaborate a bit on how you plan to use this concretely?

vchuravy:
> Can you please elaborate a bit on how you plan to use this concretely?

For projects like OpenMPI_jll that use Preferences for artifact selection we need to add the package that holds the "key=>value" to the deps, thus we added OpenMPI_jll, but when the artifact is not available we create a cache file
in which we never have done `using MPIPreferences`. Thus the cache file is not invalidated when the cache file changes. Leading to the behaviour in JuliaParallel/MPI.jl#646

giordano:
`OpenMPI_jll` already depends on `MPIPreferences` in the project file: https://github.com/JuliaBinaryWrappers/OpenMPI_jll.jl/blob/a4b7bf25afd5e5c134e8e4f9117d47c2cdf12ac2/Project.toml#L6.  I'm missing what this PR adds to that.

vchuravy:
When we cache a Package we do not care about the Project.toml, the cache file is only dependent on the code being parsed. We track what files are `included` and what Packages are pulled in with `import/using`. The cache file thus is missing a dependency edge on MPIPreferences and only MPIPreferences is invalidated when the Preference itself is changed.

giordano:
To understand: this means that we'll have `using MPIPreferences` in `src/OpenMPI_jll.jl` instead of `src/wrappers/PLATFORM.jl`?

vchuravy:
Yes exactly, we just move it up so that when there is no valid wrapper file we still add the dependency

giordano:
Ok, now I got it! :smiley:

vchuravy:
@giordano I don't think I am responsible for the 1.8 error xD

src/Dependencies.jl
giordano, line 103: Document this argument?  You may also want to add it to the other `Dependency` methods.

giordano (approved):

src/Dependencies.jl
vchuravy: Let me know if the docs are to detailed.

giordano: I think it's fine :slightly_smiling_face:  Thanks!

src/Dependencies.jl
```diff
@@ -86,14 +94,24 @@ The optional keyword argument `platforms` is a vector of `AbstractPlatform`s
...
+and in the case that no platform is available we need to be able to invalidate
+said cache. Invalidation occurs through the package that owns the `Preferences`
+data.
 """
 struct Dependency <: AbstractDependency
     pkg::PkgSpec
     build_version::Union{VersionNumber,Nothing}
     compat::String  # semver string for use in Project.toml of the JLL
     platforms::Vector{<:AbstractPlatform}
+    top_level::Bool
```
giordano, line 111: It just occurred to me that top-level dependencies can only be `RuntimeDependency`, right?  Or do you see cases where it can be a built-time dependency too?

giordano: This would make #278 largely unnecessary because those dependencies would be pulled in during the build, and so they wouldn't cause the extra warning in the first place.

vchuravy: Uhm the only case is @maleadt current adventure where the availability of a artifact might change the platform selection of the current one.

vchuravy: Maybe a more principled stance would be to.make everything that is AnyPlatform top level?

giordano: > Maybe a more principled stance would be to.make everything that is AnyPlatform top level?

Uhm, no [`iso_codes_jll`](https://github.com/JuliaBinaryWrappers/iso_codes_jll.jl) is a regular JLL which is `AnyPlatform`, I'm not sure that should be a top-level dependency.

giordano: Also, `AnyPlatform()` packages are often build-time dependencies (e.g. header-only packages), `iso_codes_jll` is an exception but we have to deal with that.

PR #278, opened 2022-10-04
[Prefix] Allow some dependencies not to have an `Artifacts.toml` file

giordano:
This clears warnings about `MPIPreferences` not having an `Artifacts.toml` file.  One possible objection I can see is that we may not want to hardcode Yggdrasil-specific stuff in this package, could be solved with an environment variable or perhaps a preference.

### DIFF
--- a/src/Prefix.jl
+++ b/src/Prefix.jl
@@ -594,6 +594,11 @@ end
 # Fallback for other types, like `AnyPlatform`.
 normalize_platform(p::AbstractPlatform) = p
 
+# Dependencies which are allow to not have an `Artifacts.toml` file.
+const NO_ARTIFACTS_ALLOW_LIST = [
+    "MPIPreferences",
+]
+
 function setup_dependencies(prefix::Prefix,
                             dependencies::Vector{PkgSpec},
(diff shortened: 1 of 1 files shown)


vchuravy:
Yeah not the biggest fan of having to manually encode these (and there might be more...), Maybe use the new top_level dependency property?

giordano:
I agree we should rather check that a dependency is a top-level one, but `setup_dependencies` takes in a `Vector{PkgSpec}`, not `Vector{AbstractDependency}` https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/dd8ee21b49f57bc288a48160ea3dbddb4bf284e3/src/Prefix.jl#L597-L600 so, without changing the signature of this function we can't possibly know whether a dependency was a top-level one or not.  However, as I mentioned in https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/275#discussion_r996320261 I believe these dependencies should be `RuntimeDependency`, rather than `Dependency`.  I think we should fix this in Yggdrasil by changing the type of the dependency, I'm going to close this PR.

PR #279, opened 2022-10-15
[Dependencies] Allow using `top_level` with `RuntimeDependency`

giordano:
Also, deprecate `top_level` with `Dependency`.  Ref: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/275#discussion_r996320261.

### DIFF
--- a/src/Dependencies.jl
+++ b/src/Dependencies.jl
@@ -65,7 +65,7 @@ Return whether `dep` is a runtime dependency or not.
 is_runtime_dependency
 
 """
-    is_top_level_dependency(dep::AbstractDependency) ->
+    is_top_level_dependency(dep::AbstractDependency) -> Bool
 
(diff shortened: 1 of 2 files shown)


src/Dependencies.jl
```diff
@@ -128,6 +128,7 @@ struct Dependency <: AbstractDependency
             end
         end
         if top_level
+            @warn("Dependency(\"$(getname(pkg))\") was defined as top-level but this is deprecated, use `RuntimeDependency` instead")
```
giordano, line 131: Strictly speaking, we could also just remove `top_level` from `Dependency` without doing a breaking release as it was never documented :innocent:

vchuravy (approved):

giordano:
Failure on nightly is https://github.com/JuliaLang/julia/issues/47180

PR #281, opened 2022-10-16
[Sources] Consider URLs starting with `git://` as Git repositories

giordano:
Basically same as https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/940, but in this package.  Fix bug reported by @Seelengrab on Slack.

### DIFF
--- a/src/Sources.jl
+++ b/src/Sources.jl
@@ -98,6 +98,9 @@ end
 DirectorySource(path::String; target::String = "", follow_symlinks::Bool=false) =
     DirectorySource(path, target, follow_symlinks)
 
+# Try to guess if a URL is a Git repository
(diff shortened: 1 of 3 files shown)


src/Sources.jl
```diff
@@ -98,6 +98,9 @@ end
 DirectorySource(path::String; target::String = "", follow_symlinks::Bool=false) =
     DirectorySource(path, target, follow_symlinks)
 
+# Try to guess if a URL is a Git repository
+isgitrepo(url::AbstractString) = endswith(url, ".git") || startswith(url, "git://")
```
giordano, line 102: I defined this function so that we can use it in https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/cf0121f82a09c8288f6cd3969b931d17e80b06aa/src/wizard/obtain_source.jl#L86, which should use the same logic

PR #283, opened 2022-10-29
Use `HistoricalStdlibVersions` to populate `Pkg.Types.STDLIBS_BY_VERSION`

giordano:


.github/workflows/ci.yml
```diff
@@ -66,6 +66,8 @@ jobs:
       - uses: julia-actions/julia-buildpkg@latest
       - name: System info
         run: julia --project=. --color=yes -e "using BinaryBuilderBase; BinaryBuilderBase.versioninfo()"
+      - name: Install HistoricalStdlibVersions
+        run: julia --project=. --color=yes -e 'using Pkg; Pkg.add(; url="https://github.com/JuliaPackaging/HistoricalStdlibVersions.jl")'
```
giordano, line 70: This is only while the package is being registered, no need to cut short the waiting time.

src/BinaryBuilderBase.jl
```diff
@@ -219,6 +220,9 @@ function __init__()
     if get(ENV, "BINARYBUILDER_USE_CCACHE", "false") == "true"
         use_ccache[] = true
     end
+
+    # Populate `Pkg.STDLIBS_BY_VERSION`
+    append!(empty!(Pkg.STDLIBS_BY_VERSION), HistoricalStdlibVersions.STDLIBS_BY_VERSION)
```
IanButterworth, line 225:
```suggestion
    append!(empty!(Pkg.Types.STDLIBS_BY_VERSION), HistoricalStdlibVersions.STDLIBS_BY_VERSION)
```

DilumAluthge (commented):
I think this should be done in the `BinaryBuilderBase.__init__()` function, right?

src/BinaryBuilderBase.jl
IanButterworth: Sorry, was a typo in the readme. Was correct in the docstring

giordano:
> I think this should be done in the `BinaryBuilderBase.__init__()` function, right?

Which is where it's done: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/759f3c04f08b7cb48db75a4fa656d5032cb6fee2/src/BinaryBuilderBase.jl#L170-L226 Or am I missing something?

src/BinaryBuilderBase.jl
```diff
@@ -219,6 +220,9 @@ function __init__()
     if get(ENV, "BINARYBUILDER_USE_CCACHE", "false") == "true"
         use_ccache[] = true
     end
+
+    # Populate `Pkg.STDLIBS_BY_VERSION`
```
IanButterworth, line 224: comment nit
```suggestion
    # Populate `Pkg.Types.STDLIBS_BY_VERSION`
```

DilumAluthge:
Ah, GitHub hid that part from me, so I couldn't tell that we were inside the `__init__` function.

DilumAluthge (approved):

giordano:
Sounds like this block https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/1aae9a7b60d3bac5cf31114e3423be6b6617289f/test/compat.jl#L29-L45 is finally gone for good.

PR #284, opened 2022-11-04
[Runner] Set also minimum SDK version when targeting macOS

giordano:
~Needs tests~

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -336,10 +336,9 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
         return flags
     end
 
-    function min_macos_version_flags(p::AbstractPlatform)
-        ver = macos_version(p)
+    function min_macos_version_flags()
         # Ask compilers to compile for a minimum macOS version, targeting that SDK.
-        return ("-mmacosx-version-min=$(ver)", "-Wl,-sdk_version,$(ver)")
+        return ("-mmacosx-version-min=\${MACOSX_DEPLOYMENT_TARGET}", "-Wl,-sdk_version,\${MACOSX_DEPLOYMENT_TARGET}")
     end
 
     function add_system_includedir(flags::Vector{String})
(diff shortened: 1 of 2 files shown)


src/Runner.jl
```diff
@@ -336,9 +336,10 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
         return flags
     end
 
-    function min_macos_version_flag(p::AbstractPlatform)
-        # Ask compilers to compile for a minimum macOS version
-        return "-mmacosx-version-min=$(macos_version(p))"
+    function min_macos_version_flags(p::AbstractPlatform)
+        ver = macos_version(p)
+        # Ask compilers to compile for a minimum macOS version, targeting that SDK.
+        return ("-mmacosx-version-min=$(ver)", "-Wl,-sdk_version,$(ver)")
```
giordano, line 342: @staticfloat what do you think about using here
```julia
("-mmacosx-version-min=\${MACOSX_DEPLOYMENT_TARGET}", "-Wl,-sdk_version,\${MACOSX_DEPLOYMENT_TARGET}")
```
?  The idea is that in the cases where we use a different SDK we can still set the same value by overriding the environment variable `MACOSX_DEPLOYMENT_TARGET`, which is otherwise set to the default macOS verson:
https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/43ba7401c899aefbb0dc57da368457e2fc54aade/src/Runner.jl#L1182

giordano: Looking at the value of `MACOSX_DEPLOYMENT_TARGET` is also what cctools linker should in principle do (even though I didn't manage to make it work): https://github.com/tpoechtrager/cctools-port/blob/634a084377ee2e2932c66459b0396edf76da2e9f/cctools/ld64/src/ld/Options.cpp#L5378-L5394.

Therefore I'm going with this option because it's more consistent (the environment variable `MACOSX_DEPLOYMENT_TARGET` is used by build systems, like CMake) and offers users an escape hatch when they use a different SDK.

PR #288, opened 2022-11-25
download all artifacts with concurrent tasks

johnnychen94:
It takes quite a long time to wait for each artifact to be downloaded if it's in a sequential manner, so I figure it might be worth adding concurrent support here.

But I'm not sure if this is implemented well. For instance, the downloading tasks can't be correctly killed if I hit Ctrl-C. Does Julia have a "standard" way of doing so? I found https://discourse.julialang.org/t/stop-terminate-a-sub-task-started-with-async/32193 but am not sure if that's okay to use.

### DIFF
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -798,7 +798,7 @@ function expand_cxxstring_abis(platform::AbstractPlatform; skip=Sys.isbsd)
         p["cxxstring_abi"] = "cxx11" #Clang only seems to generate cxx11 abi
         return [p]
     end
-                                                                                                                                 
+
     # Otherwise, generate new versions!
     map(["cxx03", "cxx11"]) do abi
         p = deepcopy(platform)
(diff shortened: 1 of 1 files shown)


src/Rootfs.jl
```diff
@@ -1014,20 +1014,39 @@ function preferred_cxxstring_abi(platform::AbstractPlatform, shard::CompilerShar
...
+    taskpool = Channel{Task}(concurrency) do ch
+        for name in keys(artifacts)
+            t = @task ensure_artifact_installed(name, artifacts[name], artifacts_toml; verbose=false)
+            put!(ch, t)
+        end
+    end
+    verbose && @info "launch $concurrency concurrent downloading tasks"
+    wait(taskpool)
+    @sync for _ in 1:concurrency
+        @async begin
```
vchuravy, line 1041: For new code I recommend against the use of `@async` and strongly prefer `Threads.@spawn`.

giordano:
Side note: `download_all_artifacts` isn't used anywhere [as far as I can tell](https://github.com/search?q=org%3AJuliaPackaging+download_all_artifacts&type=code).

Issue #179, opened 2021-11-11
LibraryProduct docstring mentions dir_paths kwarg, but it doesn't exist


marcom:
This should work according to the docstring of LibraryProduct:
`LibraryProduct("libfoo", :libfoo; dir_paths=["lib/plugins"])`

But instead it gives an error.

The undocumented variants `LibraryProduct("libfoo", :libfoo, ["lib/plugins"])` or `LibraryProduct("libfoo", :libfoo, "lib/plugins")` do work.

Should the docstring be changed?

giordano:
> Should the docstring be changed?

I think so, thanks

PR #291, opened 2023-01-13
fix `dir_paths` kwarg -> arg

t-bltg:
Fix https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/179, which I've just hit in a new recipe, reading https://docs.binarybuilder.org/stable/reference/#BinaryBuilderBase.LibraryProduct.

### DIFF
--- a/src/Products.jl
+++ b/src/Products.jl
@@ -76,15 +76,14 @@ build configuration.
 
 ---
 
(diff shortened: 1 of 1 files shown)


giordano:
Thanks!

PR #293, opened 2023-01-31
[Sources] Ban GitHub-generated archives as sources

giordano:


test/sources.jl
```diff
@@ -20,12 +25,12 @@ using JSON
     @testset "Download and setup" begin
         mktempdir() do dir
             cd(dir) do
-                as = ArchiveSource("https://github.com/ralna/ARCHDefs/archive/v2.0.3x.tar.gz", "6583e27f84338447767bbdf4335514c8836ae4ad54f5e66280307e8b57189cff")
+                as = ArchiveSource("https://ftp.gnu.org/pub/gnu/auctex/auctex-11.81.tar.gz", "8e9e84515013cb79debbbc06f789817ca57c6125905e84f17a69339ae56632c1")
```
giordano, line 28: I hope to find a ligher archive as source for this test, this is ~800 kiB, but the old one used to be ~16 kiB.

src/Sources.jl
```diff
@@ -17,6 +17,27 @@ Concrete subtypes of `AbstractSource` are:
...
+function check_github_archive(url::String)
+    # See
+    # * https://github.blog/changelog/2023-01-30-git-archive-checksums-may-change/
+    # * https://github.com/bazel-contrib/SIG-rules-authors/issues/11
+    # * https://github.com/spack/spack/issues/35250
+    # Note: according to
+    # <https://github.com/bazel-contrib/SIG-rules-authors/issues/11#issuecomment-1029861300>
+    # the `/archive/refs/tag` should be more stable, but that didn't happen in the incident
+    # on 2023-01-30, so it's unclear whether we can trust them.
+    if contains(url, r"github.com/[^/]+/[^/]+/archive/(refs/tag/)?[^/]+\.(tar\.gz|zip)$")
```
aviks, line 29: This won't match manually uploaded release tarballs, right?

giordano: Correct, those are `/download/`, like the file now used as test in https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/293/files#diff-5a745ed885ee2954c186a5c488dfb2f3d6c85e220cc63f5dc427035597ff0d1cR28

PR #294, opened 2023-01-31
[Sources] fix github url check

benlorenz:
fixup #293, the url has `tags` instead of `tag`. (Clicking on the previous urls [here](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/294/commits/6fb8417ebff6a4de771d3d27a2bf4c8979983cc1#diff-5a745ed885ee2954c186a5c488dfb2f3d6c85e220cc63f5dc427035597ff0d1cL15) in the commits view gives a 404)


giordano:
Ugh, thanks!

PR #306, opened 2023-03-21
[RootFS] Add Rust toolchain v1.68.0

### LINKED ISSUE #263: Use sparse registry for Cargo
When it will become available in a stable version of Rust we support, it'd be nice to automatically enable the sparse registry: https://blog.rust-lang.org/2022/06/22/sparse-registry-testing.html.  Updating the registry always takes an overly large fraction of the build process of Rust packages, in addition to the fact we can't cache build outputs (related issue: #186).

giordano:
Fix #263

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -1125,6 +1125,12 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
             merge!(mapping, Dict(
                 "RUSTUP_TOOLCHAIN" => "$(rust_version)-$(map_rust_target(host_platform))",
             ))
+            if rust_version ≥ v"1.68.0"
+                merge!(mapping, Dict(
+                    # Use sparse crates.io registry for quicker updates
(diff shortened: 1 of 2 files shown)


src/Runner.jl
```diff
@@ -1125,6 +1125,12 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
             merge!(mapping, Dict(
                 "RUSTUP_TOOLCHAIN" => "$(rust_version)-$(map_rust_target(host_platform))",
             ))
+            if rust_version ≥ v"1.68.0"
+                merge!(mapping, Dict(
+                    # Use sparse crates.io registry for quicker updates
+                    "REGISTRIES_CRATES_IO_PROTOCOL" => "sparse",
```
giordano, line 1131: This doesn't seem to be working as expected.... :sob: :sob: :sob:

giordano: It turns out the name of the environment variable was misspelled in https://github.com/rust-lang/cargo/pull/11224#issue-1406777084 (see edit history of the message...)

PR #308, opened 2023-04-24
Add defines for gcc and clang that declare we are Windows 10

marcom:
Tests seem to be failing at the moment on master for an unrelated reason.

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -334,12 +334,7 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
             end
         end
         if Sys.iswindows(p)
-            # Declare that we are Windows 10 (0x0A00)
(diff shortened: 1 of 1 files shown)


src/Runner.jl
```diff
@@ -333,6 +333,14 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
...
             end
         end
+        if Sys.iswindows(p)
+            # Declare that we are Windows 10 (0x0A00)
+            # https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=msvc-170
+            append!(flags, [
+                "-DWINVER=0x0A00",
+                "-D_WIN32_WINNT=0x0A00",
+            ])
+        end
```
giordano, line 343: I'm not very happy about duplication, it's redundant and error-prone.  Define a function (`windows_cflags`?) and use it in the two functions.

PR #309, opened 2023-04-24
Create `test/Project.toml` environment, with compat bounds

giordano:


giordano:
Failure on nightly is https://github.com/JuliaLang/julia/issues/49491

PR #314, opened 2023-06-24
[Dependencies] Require `compat` argument for runtime dependencies

giordano:


giordano:
Uhm, a problem with requiring compat everywhere is stdlibs, which really don't like enforcing version numbers, and when the compat is set the build version is inferred from that automatically.  Need to come up with a good idea for stdlibs.

PR #312, opened 2023-06-11
Update Artifacts.toml for FreeBSD 13.2

ararslan:
https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/166 for the modern era. Changes to Artifacts.toml generated with https://github.com/JuliaPackaging/Yggdrasil/pull/6865.

To do:
- [x] PlatformSupport
- [ ] GCCBootstrap@4
- [ ] GCCBootstrap@5
- [ ] GCCBootstrap@6
- [ ] GCCBootstrap@7
- [ ] GCCBootstrap@8
- [ ] GCCBootstrap@9
- [ ] GCCBootstrap@10
- [ ] GCCBootstrap@11
- [ ] GCCBootstrap@11-IainS
- [ ] GCCBootstrap@12
- [ ] GCCBootstrap@12-IainS
- [ ] Determine whether the 12.2 artifacts should be removed from the file as was done when going from 11 to 12
- [ ] Bump version

PR #315, opened 2023-07-11
Remove GCCBootstrap artifacts for FreeBSD 12.2

ararslan:
The old GCCBootstrap artifacts for FreeBSD 12.2 seem to be picked up by Yggdrasil CI rather than the newer 13.2 artifacts, but the 13.2 artifacts are the ones that should be used.

Note that the 12.2 artifacts for PlatformSupport are left untouched because in that case the 12.2 and 13.2 artifacts correspond to distinct versions of PlatformSupport.

The minor version is bumped here, even though this is technically breaking for users on FreeBSD 12. However, FreeBSD 12 is past EOL so I think it's reasonable to make it a minor rather than major change.


giordano (approved):

ararslan:
Looks like the CI failures are the same as those observed in #312, which is a known issue.

PR #322, opened 2023-09-20
Store `url` in SetupSource

Keno:
Otherwise the wizard needs to keep a separate 1-1 mapping of the urls, which is somewhat silly.


giordano (commented):
It'd be good to bump the minor version of this package so that we can require that in https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1293

giordano:
The more I look at this, the more I'm convinced it's a breaking change.  That's fine, [I don't see any users apart from BinaryBuilder](https://juliahub.com/ui/Search?type=symbols&q=SetupSource&u=use), but we still need to change the major version number not to let people end up with incompatible versions.

PR #323, opened 2023-09-20
Add GCC 13.2 shards to BB

gbaraldi:


gbaraldi:
Is this mergeable?

giordano:
I'd still like to see the corresponding PR in Yggdrasil, I don't like merging PRs here if we don't have the companion too

giordano:
BTW, I think we should first get #318 merged.  @staticfloat any reason why that hasn't been done yet?

staticfloat:
Nope, just crazy busy.  Sorry.

gbaraldi:
Do I need to rebuild then?

giordano:
If you were after https://github.com/JuliaPackaging/Yggdrasil/pull/7136 on Yggdrasil, you should be good.

PR #324, opened 2023-09-20
Add clang-16

### LINKED ISSUE #248: Clang wrapper for Windows isn't working

### LINKED ISSUE #290: Clang wrapper for Linux and Windows should set `_GLIBCXX_USE_CXX11_ABI`
At the moment https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/f782dc5d327dfa21cc26ae582e4d6424f7df8bc5/src/Runner.jl#L462-L467 is done only for GCC, but it should be done also for Clang on Linux and Windows, since in that case we use libstdc+.

gbaraldi:
This fixes #290 and fixes #248

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -322,6 +322,11 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
             "--sysroot=/opt/$(aatriplet(p))/$(aatriplet(p))/sys-root",
         ])
         if !Sys.isbsd(p)
+            if cxxstring_abi(p) == "cxx11"
+                push!(flags, "-D_GLIBCXX_USE_CXX11_ABI=1")
+            elseif cxxstring_abi(p) == "cxx03"
+                push!(flags, "-D_GLIBCXX_USE_CXX11_ABI=0")
+            end
             if iscxx
                 append!(flags, [
                     # Link with libstdc++ when compiling c++ on non-BSDs
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -141,6 +141,15 @@ end
 prefer_clang(p::AbstractPlatform) =
     Sys.isbsd(p) || sanitize(p) in ("memory", "memory_origins", "address")
 
+# Add the string ABI define
+function add_cxx_abi(p::AbstractPlatform, flags::Vector{String})
+    if cxxstring_abi(p) == "cxx11"
+        push!(flags, "-D_GLIBCXX_USE_CXX11_ABI=1")
+    elseif cxxstring_abi(p) == "cxx03"
+        push!(flags, "-D_GLIBCXX_USE_CXX11_ABI=0")
+    end
+end
+
 """
     generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::AbstractString,
                                 host_platform::AbstractPlatform = $(repr(default_host_platform)),
@@ -162,6 +171,7 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
                                      allow_unsafe_flags::Bool = false,
                                      lock_microarchitecture::Bool = true,
                                      bootstrap::Bool = !isempty(bootstrap_list),
+                                     gcc_version::Union{Nothing,VersionNumber}=nothing,
                                      )
     # Wipe that directory out, in case it already had compiler wrappers
     rm(bin_path; recursive=true, force=true)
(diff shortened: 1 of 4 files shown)


src/Runner.jl
```diff
@@ -415,6 +421,18 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
...
+        if !Sys.isbsd(p)
+            gcc_version, llvm_version = select_compiler_versions(p, compilers)
+            append!(flags, String["-I/opt/$(aatriplet(p))/$(aatriplet(p))/include/c++/$(gcc_version)",
+                                  "-I/opt/$(aatriplet(p))/$(aatriplet(p))/include/c++/$(gcc_version)/$(aatriplet(p))",
+                                  "-I/opt/$(aatriplet(p))/$(aatriplet(p))/include/c++/$(gcc_version)/backward",
+                                  "-I/opt/$(aatriplet(p))/lib/gcc/$(aatriplet(p))/$(gcc_version)/include",
+                                  "-I/opt/$(aatriplet(p))/lib/gcc/$(aatriplet(p))/$(gcc_version)/include-fixed",
+                                  "-I/opt/$(aatriplet(p))/$(aatriplet(p))/include",
+                                  "-I/opt/$(aatriplet(p))/$(aatriplet(p))/sys-root/include"])
+        end
```
giordano, line 434: Why this instead of `--gcc-toolchain`?  Did you see #247 and #248?  One problem of this approach is also that this triggers `-Werror` failures in case some paths are unused, which isn't great.

giordano (commented):
It'd be good to see the corresponding PR to Yggdrasil as well.

src/Runner.jl
```diff
@@ -322,17 +322,23 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
...
         if !Sys.isbsd(p)
-            append!(flags, [
-                # Find GCC toolchain here (for things like libgcc_s)
-                "--gcc-toolchain=/opt/$(aatriplet(p))"
-            ])
+            if cxxstring_abi(p) == "cxx11"
+                push!(flags, "-D_GLIBCXX_USE_CXX11_ABI=1")
+            elseif cxxstring_abi(p) == "cxx03"
+                push!(flags, "-D_GLIBCXX_USE_CXX11_ABI=0")
+            end
```
giordano, line 329: Maybe define a function to be used both in GCC and Clang, to avoid duplication?

src/Runner.jl
gbaraldi: gcc-toolchain wasn't reliable for me when testing, our gcc install is a bit weird because we set a sysroot that isn't the usual one and clang isn't able to find stuff. Specially in mingw.
About the warning, given that it didn't complain when compiling very simple C code I would be surprised if it did for more complicated ones?

giordano: I'm pretty sure I've seen that warning in the wild.

gbaraldi: I discovered why it might complain, we should use -isystem here. Those if unused do not get warnings.

giordano: Ah, that's probably a good idea here

gbaraldi:
Given that this successfully built LLVM I think we can say it's working

giordano:
Thanks!

src/Runner.jl
```diff
@@ -322,17 +332,17 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
             "--sysroot=/opt/$(aatriplet(p))/$(aatriplet(p))/sys-root",
         ])
         if !Sys.isbsd(p)
-            append!(flags, [
-                # Find GCC toolchain here (for things like libgcc_s)
-                "--gcc-toolchain=/opt/$(aatriplet(p))"
-            ])
```
giordano, line 328: Why was this removed?

PR #330, opened 2023-09-28
Add new rebuilt mingw shards

gbaraldi:
This bumps the mingw runtime to the 11.0.1 version. mingw + mscvrt doesn't have version constraints because it's a statically linked runtime that links to  msvcrt.dll which is always there


giordano:
We have some pesky tests to updated: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/2338cb3153ff7b9ffe4be0173a3199c04ac056e9/test/rootfs.jl#L180 🙂

gbaraldi:
I think this looks better

PR #332, opened 2023-10-06
Fix old clang/llvm version that broke after LLVM16

gbaraldi:
While this doesn't fix the issues with LLVM16Bootstrap, this should restore the behaviour of the older versions. Also added a test on non default LLVM/GCC versions.

### DIFF
--- a/test/runners.jl
+++ b/test/runners.jl
@@ -124,13 +124,9 @@ end
     end
 
     # This tests only that compilers for all platforms can build and link simple C code
-    @testset "Compilation - $(platform) - $(compiler)" for platform in platforms, compiler in ("cc", "gcc", "clang"), version in (nothing, v"13")
+    @testset "Compilation - $(platform) - $(compiler)" for platform in platforms, compiler in ("cc", "gcc", "clang")
         mktempdir() do dir
-            if isnothing(version)
-                ur = preferred_runner()(dir; platform=platform)
-            else
-                ur = preferred_runner()(dir; platform=platform, preferred_gcc_version=version, preferred_llvm_version=version)
-            end
+            ur = preferred_runner()(dir; platform=platform)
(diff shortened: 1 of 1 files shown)


giordano:
Can you cherry pick #321 and see if that solves the test failure?  This is also going to make the artifacts uncacheable.  Triple sigh.

test/runners.jl
```diff
@@ -124,9 +124,13 @@ end
     end
 
     # This tests only that compilers for all platforms can build and link simple C code
-    @testset "Compilation - $(platform) - $(compiler)" for platform in platforms, compiler in ("cc", "gcc", "clang")
+    @testset "Compilation - $(platform) - $(compiler)" for platform in platforms, compiler in ("cc", "gcc", "clang"), version in (nothing, v"13")
```
giordano, line 127: Sadly I don't think it's feasible to do this test with the current CI setup.

gbaraldi: That's a shame. Maybe we should setup this in buildkite? Maybe sharing the cache with Yggy?

giordano: Yggdrasil doesn't have much cache, it's evicted every other day

giordano (commented):
To be clear, with this PR Clang will still be broken again for Windows for version 13-, right?  Not a big deal since this is broken anyway at the moment.

The PR looks good to me.  Maybe open an issue to remind us to extend the tests when we'll have a better CI provision?  I don't think it's worth blocking this PR for unrelated infrastructure problems and I tested this makes OpenBLAS build successfully in https://github.com/JuliaPackaging/Yggdrasil/pull/7358#issuecomment-1752131138

gbaraldi:
It might still work, I'm not sure.

gbaraldi:
Running locally it seems to work.

PR #340, opened 2023-10-27
Default clang to use lld on specific conditions

gbaraldi:
This switches clang to use lld when the clang version is 16 or above and the GCC version is 5 or above, It passes the commented out version of the test locally but given the issues we are seeing I'leaving it commented out for now.

The main goal of this is so that clang doesn't complain about very new linker features that the shipped binutils might not have.

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -187,6 +187,11 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
 
     target = aatriplet(platform)
     host_target = aatriplet(host_platform)
+    if (!isnothing(gcc_version) && !isnothing(clang_version) && clang_version >= v"16" && gcc_version >= v"5")
+        clang_use_lld = true
+    else
+        clang_use_lld = false
+    end
 
     function wrapper(io::IO,
                      prog::String;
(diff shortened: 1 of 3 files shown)


src/Runner.jl
```diff
@@ -187,6 +187,11 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
 
     target = aatriplet(platform)
     host_target = aatriplet(host_platform)
+    if (!isnothing(gcc_version) && !isnothing(clang_version) && clang_version >= v"16" && gcc_version >= v"5")
+        clang_use_lld = true
+    else
+        clang_use_lld = false
+    end
```
giordano, line 194:
```suggestion
    clang_use_lld = (!isnothing(gcc_version) && !isnothing(clang_version) && clang_version >= v"16" && gcc_version >= v"5")
```

test/rootfs.jl
```diff
@@ -223,7 +223,6 @@ end
             clang = read(joinpath(platform_bin_dir, "clang"), String)
             # Check link flags
             @test occursin("-L/opt/$(triplet(platform))/$(triplet(platform))/lib", clang)
-            @test occursin("fuse-ld=$(triplet(platform))", clang)
```
giordano, line 226: Can we check something else here?

gbaraldi: We don't set any guaranteed flags except that one in freebsd. But I will add another that checks the gcc install stuff in linux.

gbaraldi:
Bump!

PR #342, opened 2023-10-30
Add specific test for lld usage in windows

gbaraldi:
I wasn't really satisfied with the coverage I had, so this is a more specific test.

### DIFF
--- a/test/runners.jl
+++ b/test/runners.jl
@@ -428,6 +428,39 @@ end
             @test contains(lines[end], r"^ +sdk 10\.14$")
         end
     end
+
+    @testset "Check that we use lld successfully in windows" begin
+        mktempdir() do dir
+            ur = preferred_runner()(dir; platform=Platform("x86_64", "windows"), preferred_gcc_version=v"5")
+            iobuff = IOBuffer()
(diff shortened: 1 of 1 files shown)


test/runners.jl
```diff
@@ -428,6 +428,39 @@ end
             @test contains(lines[end], r"^ +sdk 10\.14$")
         end
     end
+
+    @testset "Check that we use lld successfully in windows" begin
+        mktempdir() do dir
+            ur = preferred_runner()(dir; platform=Platform("x86_64", "windows"), preferred_gcc_version=v"5")
+            iobuff = IOBuffer()
+            test_c = """
+            #include <stdlib.h>
```
giordano, line 437: Out of curiosity, is stdlib necessary for reproducing some bugs you observed before? I'm asking just because it doesn't look like you're using anything from that header directly?

giordano (approved):

test/runners.jl
gbaraldi: So doing the clang+windows stuff, the tests passed but it was still broken because it didn't look for any headers.

Issue #349, opened 2023-11-06
`cc -fuse-ld=lld` is broken due to GCC being unable to find `ld.lld`


topolarity:
```terminal
sandbox:${WORKSPACE}/srcdir # cc -fuse-ld=lld
collect2: fatal error: cannot find ‘ld’
compilation terminated.
```

Note that the error message is something of a lie - it's `ld.lld` that it cannot find not `ld`.

(You can see that if you do `strace -f cc -fuse-ld=lld 2>&1 | grep "ld\""` which will show the files attempted to be opened by GCC)

topolarity:
Seen when putting together https://github.com/JuliaPackaging/Yggdrasil/pull/7621

PR #350, opened 2023-11-06
Fixup lld code even more and make sure gcc can use it

gbaraldi:
Fixes https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/349. Though -fuse-ld=lld is only available to GCC9 and up IIRC.
Also supercedes https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/342

### DIFF
--- a/test/runners.jl
+++ b/test/runners.jl
@@ -471,7 +471,7 @@ end
             test_script = """
                 set -e
                 echo '$(test_c)' > test.c
-                cc -Werror -shared test.c -fuse-ld=lld -o libtest.\${dlext}
+                gcc -Werror -shared test.c -fuse-ld=lld -o libtest.\${dlext}
                 """
             cmd = `/bin/bash -c "$(test_script)"`
             @test run(ur, cmd, iobuff)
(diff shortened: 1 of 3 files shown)


test/runners.jl
```diff
@@ -428,6 +428,68 @@ end
...
+            test_c = """
+            #include <stdlib.h>
+            int test(void) {
+                return 0;
+            }
+            """
+            test_script = """
+                set -e
+                echo '$(test_c)' > test.c
+                cc -Werror -shared test.c -fuse-ld=lld -o libtest.\${dlext}
```
giordano, line 474: Might want to be explicit in this test about using `gcc` instead of whatever is `cc` (in case we'll eventually change the default compiler)
```suggestion
                gcc -Werror -shared test.c -fuse-ld=lld -o libtest.\${dlext}
```

test/runners.jl
```diff
@@ -428,6 +428,68 @@ end
...
+                """
+            cmd = `/bin/bash -c "$(test_script)"`
+            @test run(ur, cmd, iobuff)
+
+            test_script = """
+            set -e
+            echo '$(test_c)' > test.c
+            cc -Werror -v -shared test.c -fuse-ld=lld -o libtest.\${dlext}
```
giordano, line 482: Ditto
```suggestion
            gcc -Werror -v -shared test.c -fuse-ld=lld -o libtest.\${dlext}
```

PR #352, opened 2023-11-10
Babysit meson so it doesn't do the wrong thing

gbaraldi:
Meson tries really hard to do it wrong.
Also note for the future `-fuse-ld` behaves very differently between `gcc` and `clang`. GCC only wants a linker name, while clang understands a full path and one with platform specific handling. Also gcc calls it `bfd` while clang wants `ld`

### DIFF
--- a/src/BuildToolchains.jl
+++ b/src/BuildToolchains.jl
@@ -31,11 +31,17 @@ function cmake_os(p::AbstractPlatform)
     end
 end
 
-lld_str(p::AbstractPlatform) = Sys.isapple(p) ? "ld64.lld" : "ld.lld"
-function linker_string(p::AbstractPlatform, clang_use_lld)
+lld_string(p::AbstractPlatform) = Sys.isapple(p) ? "ld64.lld" : "ld.lld"
+function linker_string(bt::CMake{:gcc}, p::AbstractPlatform, clang_use_lld)
     target = triplet(p)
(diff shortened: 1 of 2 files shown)


src/BuildToolchains.jl
```diff
@@ -179,9 +197,9 @@ function toolchain_file(bt::Meson, p::AbstractPlatform, envs::Dict{String,String
     fortran = '/opt/bin/$(target)/$(aatarget)-$(fortran_compiler(bt))'
     objc = '/opt/bin/$(target)/$(aatarget)-cc'
     ar = '/opt/bin/$(target)/$(aatarget)-ar'
-    ld = '$(linker_string(p, clang_use_lld))'
-    cpp_ld = '$(linker_string(p, clang_use_lld))'
-    c_ld = '$(linker_string(p, clang_use_lld))'
+    ld = '$(linker_string(bt, p, clang_use_lld))'
+    cpp_ld = '$(linker_string(bt, p, clang_use_lld))'
+    c_ld = '$(linker_string(bt, p, clang_use_lld))'
```
giordano, line 202: I thought here you needed the basename of the linker?

gbaraldi: It's not even basename. GCC wants an actual name. While clang understands a full path. So this dispatches on the compiler now

giordano (approved):

PR #357, opened 2023-12-28
[Runner] Make `clang_use_lld` a keyword argument

giordano:
I need to do more testing (hence opening as draft), but this might be sufficient to let users set `clang_use_lld` as keyword argument to `BinaryBuilder.build_tarballs`, so that they can easily opt-out of it.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "BinaryBuilderBase"
 uuid = "7f725544-6523-48cd-82d1-3fa08ff4056e"
 authors = ["Elliot Saba <staticfloat@gmail.com>"]
-version = "1.28.0"
+version = "1.29.0"
 
 [deps]
 Bzip2_jll = "6e34b625-4abd-537c-b88f-471c36dfa7a0"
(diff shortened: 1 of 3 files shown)


giordano:
Ok, I confirm the change I proposed in this PR is indeed enough here in BinaryBuilderBase, but we need another small change in BinaryBuilder so that we can use the new keyword argument in `BinaryBuilder.build_tarballs`.

giordano:
The test failure on nightly is unrelated to this PR and it's caused by this new warning being printed by Pkg:
```
┌ Warning: Circular dependency detected. Precompilation will be skipped for:
│   Downloads [f43a241f-c20a-4ad4-852c-f6b1247861c6]
│   Pkg [44cfe95a-1eb2-52ea-b672-e2afdf69b78f]
│   p7zip_jll [3f19e933-33d8-53b3-aaab-bd5110c3b7a0]
│   MozillaCACerts_jll [14a3606d-f60d-562e-9121-12d972cd8159]
│   LibCURL [b27032c2-a3e7-50c8-80cd-2d36dbcbfd21]
└ @ Pkg.API ~/repo/julia/usr/share/julia/stdlib/v1.11/Pkg/src/API.jl:1260
```
[This CI job](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/actions/runs/7034019530) back on November 29th 2023 was successful.

PR #360, opened 2023-12-29
[CI] Make `System info` step issue code coverage info

giordano:


giordano:
This PR has the nice side effect of avoiding recompilation of a bunch of packages in the "Run tests" step, now that we're using the same options to start julia in the previous step, saving about 30 seconds in all jobs (at least in Julia v1.9+).

PR #363, opened 2024-01-01
Update Objectfile to 0.4.0

eschnett:


test/Project.toml
```diff
@@ -13,4 +13,4 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [compat]
 ConstructionBase = "1.5.1"
-ObjectFile = "0.3.6"
+ObjectFile = "0.3.6, 0.4"
```
giordano, line 16: Does this really work with both ObjectFile 0.3.6 _and_ 0.4?  If I try to iterate over a handle with ObjectFile 0.3.6 I get
```
julia> readmeta("libjulia.dylib") do ohs
           foreach(ohs) do oh
               symbols = symbol_name.(Symbols(oh))
           end
       end
ERROR: MethodError: no method matching iterate(::MachOHandle{IOStream})
```
At this point we can just set compatibility with 0.4 and move on.

test/dependencies.jl
```diff
@@ -184,7 +184,7 @@ end
             # Since Julia v1.9 Zlib_jll doesn't have logs directory at all
             @static if VERSION < v"1.9-DEV"
                 zlib_log_files = if os(platform) == "macos"
-                    ["Zlib.log.gz", "fix_identity_mismatch_libz.1.2.11.dylib.log.gz", "ldid_libz.1.2.11.dylib.log.gz"]
+                    ["Zlib.log.gz", "fix_identity_mismatch_libz.1.2.12.dylib.log.gz", "ldid_libz.1.2.12.dylib.log.gz"]
```
giordano, line 187: Why did you change this?  How did it work before?

test/Project.toml
eschnett: Good catch. I somehow thought that scalars would automatically be an iterator of length 1, but of course that's only true for numbers.

test/dependencies.jl
eschnett: I changed it because this test failed. I don't know how it worked before – I assume the test was written when `libz @1.2.11` was current.

eschnett:
**ping**?

test/dependencies.jl
giordano: I was very confused about why this didn't fail in CI (nor it failed for me recently), but this code is run only on macOS (and running Docker in macOS in CI is a mess: #133).  I guess the change is fine then.

PR #368, opened 2024-02-28
Go: New version 1.22.0

eschnett:


eschnett:
The CI error is caused by `StyledString` and seems unrelated.

PR #370, opened 2024-03-28
Update Artifacts.toml

eschnett:
New entries for Rootfs.v2024.3.28.
This updates Meson to version `1.4.0`.


giordano:
This isn't working.

giordano:
It's missing the `download` section, to fetch the artifacts.

eschnett:
I thought these would be generated automatically. The PR here includes the whole file as generated by BinaryBuilder. I think the actual upload failed because the artifacts already existed in the remote repo. Maybe this accidentally deleted the `download` section.

I'll try again tomorrow. When the file names change the uploads should succeed, maybe this will create the `download` section.

PR #373, opened 2024-04-03
[Sources] Explicit conversion to work around breaking change in ProgressMeter

giordano:
This was broken by https://github.com/timholy/ProgressMeter.jl/pull/304.


MarcMush:
`update!(p; max_steps)` should be the way to go

it's not really documented either but I would consider it more like a feature whereas `p.n` is an internal value

PR #375, opened 2024-04-09
Support riscv64 platform

eschnett:


src/Runner.jl
```diff
@@ -1288,7 +1290,7 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
             mapping["GNU_LIBC_VERSION"] = "glibc 2.12.2"
         elseif arch(platform) in ("armv7l", "aarch64")
             mapping["GNU_LIBC_VERSION"] = "glibc 2.19"
-        elseif arch(platform) === "powerpc64le"
+        elseif arch(platform) in ("powerpc64le", "riscv64")
```
giordano, line 1293: I'm not really sure this makes sense: [glibc 2.17 is from December 2012](https://sourceware.org/glibc/wiki/Glibc%20Timeline), [RISC-V was introduced in 2014](https://www2.eecs.berkeley.edu/Pubs/TechRpts/2014/EECS-2014-146.pdf), we need to find a version of glibc which actually supports RISC-V.

eschnett: Using 2.35 now.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "BinaryBuilderBase"
 uuid = "7f725544-6523-48cd-82d1-3fa08ff4056e"
 authors = ["Elliot Saba <staticfloat@gmail.com>"]
-version = "1.32.0"
+version = "1.34.0"
```
giordano, line 4: Any reason for jumping 1.33?

src/Rootfs.jl
```diff
@@ -631,9 +644,11 @@ function choose_shards(p::AbstractPlatform;
         )
 
         # We _always_ need Rootfs and PlatformSupport for our target, at least
+        # TODO: Remove this again
```
giordano, line 647: Reminder to remove this.

src/Runner.jl
```diff
@@ -1284,10 +1286,12 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
     if libc(platform) === "glibc"
         # This should be kept in sync with the version of glibc used in
         # https://github.com/JuliaPackaging/Yggdrasil/blob/master/0_RootFS/gcc_sources.jl
-        if arch(platform) in ("armv7l", "aarch64")
+        if arch(platform) in ["armv7l", "aarch64"]
```
giordano, line 1289: Why this change? Tuples are slightly more efficient as they don't allocate memory at all.

src/Runner.jl
```diff
@@ -1284,10 +1286,12 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
...
             mapping["GNU_LIBC_VERSION"] = "glibc 2.19"
-        elseif arch(platform) in ("x86_64", "i686",  "powerpc64le")
+        elseif arch(platform) in ["x86_64", "i686",  "powerpc64le"]
```
giordano, line 1291: Keep tuples also here.

src/Runner.jl
```diff
@@ -1284,10 +1286,12 @@ function platform_envs(platform::AbstractPlatform, src_name::AbstractString;
...
             mapping["GNU_LIBC_VERSION"] = "glibc 2.17"
+        elseif arch(platform) in ["riscv64"]
```
giordano, line 1293:
```suggestion
        elseif arch(platform) in ("riscv64",)
```

src/Rootfs.jl
```diff
@@ -471,6 +472,12 @@ function gcc_version(p::AbstractPlatform,GCC_builds::Vector{GCCBuild},
         GCC_builds = filter(b -> getversion(b) ≥ v"7", GCC_builds)
     end
 
+    # We only have GCC 12 or newer for RISC-V.
+    # (This could be changed down to GCC 7.1.)
```
giordano, line 476: I think it's fine to have only recent toolchains for recently introduced platforms, it reduces the maintenance burden.

src/Rootfs.jl
```diff
@@ -517,6 +524,12 @@ function gcc_version(p::AbstractPlatform,GCC_builds::Vector{GCCBuild},
 end
 
 function llvm_version(p::AbstractPlatform, LLVM_builds::Vector{LLVMBuild})
+    # # We only have LLVM 18 or newer for RISC-V.
+    # # (This could be changed down to LLVM 6.0.)
+    # if arch(p) == "riscv64"
+    #     LLVM_builds = filter(b -> getversion(b) ≥ v"6", LLVM_builds)
```
giordano, line 530: Should this be uncommented and filter llvm 18+?

src/Rootfs.jl
```diff
@@ -631,9 +644,11 @@ function choose_shards(p::AbstractPlatform;
         )
 
         # We _always_ need Rootfs and PlatformSupport for our target, at least
+        # TODO: Remove this again
+        ps_build_new = arch(p) == "riscv64" ? v"2024.12.21" : ps_build
         append!(shards, [
             find_shard("Rootfs", rootfs_build, archive_type),
-            find_shard("PlatformSupport", ps_build, archive_type; target=p)
+            find_shard("PlatformSupport", ps_build_new, archive_type; target=p)
```
giordano, line 651: Reminder to change back also this one.

Project.toml
eschnett: no

src/Rootfs.jl
eschnett: Replaced by an `if` statement.

src/Runner.jl
eschnett: Arrays are used in this file in so many places... Really, this kind of efficiency isn't needed here.

Changed to tuples.

src/Runner.jl
eschnett: Done.

src/Runner.jl
eschnett: Done.

src/Rootfs.jl
eschnett: It seems riscv64 support was introduced in LLVM 3.9 (I didn't test this) so we should be good without this.

src/Rootfs.jl
eschnett: Ok.

eschnett:
Question: Is this
```
    # We have experimental support for some platforms, allow easily including them
    if experimental
        append!(standard_platforms, [
            Platform("riscv64", "linux"),
            Platform("riscv64", "linux"; libc="musl"),
        ])
    end
```
reasonable? That is, do we tag this platform as "experimental", and do we include a `musl` build?

For context: Musl is much easier to install than Glibc, and "historically" speaking, musl was working for me before glibc did. We should acknowledge this.

giordano:
I don't think we're going to use "experimental" anymore, in hindsight it was a bit cumbersome.

About musl, we could perhaps have it, but in practice Julia has lots of troubles running on musl (we can't even compile it for i686, and other architectures like x86_64 crash continuously) that I'm not so sure it's worth the effort adding it.

src/Rootfs.jl
giordano: I thought we had to enable the targets in the LLVM recipe, but I see we have all the targets enabled by default:
```console
sandbox:${WORKSPACE} # /opt/x86_64-linux-musl/bin/clang -print-targets

  Registered Targets:
    aarch64     - AArch64 (little endian)
    aarch64_32  - AArch64 (little endian ILP32)
    aarch64_be  - AArch64 (big endian)
    amdgcn      - AMD GCN GPUs
    arm         - ARM
    arm64       - ARM64 (little endian)
    arm64_32    - ARM64 (little endian ILP32)
    armeb       - ARM (big endian)
    avr         - Atmel AVR Microcontroller
    bpf         - BPF (host endian)
    bpfeb       - BPF (big endian)
    bpfel       - BPF (little endian)
    hexagon     - Hexagon
    lanai       - Lanai
    loongarch32 - 32-bit LoongArch
    loongarch64 - 64-bit LoongArch
    mips        - MIPS (32-bit big endian)
    mips64      - MIPS (64-bit big endian)
    mips64el    - MIPS (64-bit little endian)
    mipsel      - MIPS (32-bit little endian)
    msp430      - MSP430 [experimental]
    nvptx       - NVIDIA PTX 32-bit
    nvptx64     - NVIDIA PTX 64-bit
    ppc32       - PowerPC 32
    ppc32le     - PowerPC 32 LE
    ppc64       - PowerPC 64
    ppc64le     - PowerPC 64 LE
    r600        - AMD GPUs HD2XXX-HD6XXX
    riscv32     - 32-bit RISC-V
    riscv64     - 64-bit RISC-V
    sparc       - Sparc
    sparcel     - Sparc LE
    sparcv9     - Sparc V9
    systemz     - SystemZ
    thumb       - Thumb
    thumbeb     - Thumb (big endian)
    ve          - VE
    wasm32      - WebAssembly 32-bit
    wasm64      - WebAssembly 64-bit
    x86         - 32-bit X86: Pentium-Pro and above
    x86-64      - 64-bit X86: EM64T and AMD64
    xcore       - XCore
```

But then I'm missing why this PR is updating the LLVMBootstrap artifact for llvm 18 (which is why I thought we had to manually enable the targets)

src/Platforms.jl
```diff
@@ -158,7 +158,10 @@ const ARCHITECTURE_FLAGS = Dict(
             #"power9"  => ["-mcpu=power9", "-mtune=power9"],
             # Eventually, we'll support power10, once we have compilers that support it.
             #"power10" => ["-mcpu=power10", "-mtune=power10"],
-        )
+        ),
+        "riscv64" => OrderedDict(
+            "riscv64" => [],
```
giordano, line 163: We probably need to come up with a basenline ISA/ABI to target here.  `-march=rv64g -mabi=lp64d`?

eschnett: @maleadt used `rv64gc_zba_zbb ` in https://github.com/JuliaLang/julia/pull/56105.

maleadt: That was my local build for my VisionFive2. For the more generic builds at https://github.com/maleadt/julia/releases, I used `rv64gc` (aka. `generic-rv64,i,m,a,f,d,zicsr,zifencei,c` for `JULIA_CPU_TARGET`). I haven't put a lot of thought in it, but `rv64gc` seems like a good baseline, with support for other profiles (e.g. `rva22`) through platform augmentation tags.

eschnett: Using your suggestion.

src/Rootfs.jl
eschnett: That's probably not necessary. I'm not quite sure about the new artifacts – please review them, quite a few might not be necessary.

giordano: Did you change anything locally relative to llvm? I don't see anything in https://github.com/JuliaPackaging/Yggdrasil/pull/8468

eschnett: I don't think I did. I got confused by GCC (which requires a new build for every architecture) and LLVM (which doesn't).

giordano:
I can confirm simple executables built on this PR with
```console
$ julia +1.7 -e 'using BinaryBuilderBase; BinaryBuilderBase.runshell(Platform("riscv64", "linux"))'
sandbox:${WORKSPACE} # echo -e '#include <stdio.h>\nint main(void) { printf("Hello world!\\n"); return 0; }' | gcc -x c - -o hello-gcc
sandbox:${WORKSPACE} # echo -e '#include <stdio.h>\nint main(void) { printf("Hello world!\\n"); return 0; }' | clang -x c - -o hello-clang
```
both work successfully on RISC-V:
```console
$ ./hello-clang 
Hello world!
$ ./hello-gcc
Hello world!
```

:partying_face:

src/Rootfs.jl
giordano: I'd suggest reverting the new LLVMBootstrap shard then.

eschnett: Done.

src/Rootfs.jl
```diff
@@ -631,9 +639,11 @@ function choose_shards(p::AbstractPlatform;
         )
 
         # We _always_ need Rootfs and PlatformSupport for our target, at least
+        # We don't have old platform support for riscv64. Remove this once all platfor support is aligned in time.
```
giordano, line 642: Is this still necessary after https://github.com/JuliaPackaging/BinaryBuilderBase.jl/commit/dafa99380c04a0de80e1b478a3517839bc494dc1, since that's the only relevant version?

eschnett: Yes. BinaryBuilderBase chooses `PlatformSupport` by date, and it wants specifically `v"2024.08.10"`. That's different from `Rootfs` where it chooses the newest available one.

giordano:
I think this is basically good to go from my point of view as soon as https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1355#discussion_r1894982076 is worked out (I don't have strong opinions about that).

eschnett:
We haven't found any evidence for page sizes other than 4 kB, so let's go with the current state.

src/Platforms.jl
giordano: According to https://wiki.gentoo.org/wiki/RISC-V_Multilib `-march=rv64gc -mabi=lp64d` is the default ISA/ABI used by GCC, so it sounds like a good default baseline. I also found https://www.sifive.com/blog/all-aboard-part-1-compiler-args to be an interesting read about compiler flags for RISC-V (all those extensions are a mess)

PR #377, opened 2024-05-07
Added fix for DockerRunner on macOS >= 13 wrt. xattrs

stemann:
Fixes DockerRunner for macOS >= 13 wrt. xattrs added by macOS which Docker does not like.


stemann:
Similar issue: https://github.com/moby/moby/issues/47517

dpo:
Would you please register a new version of BinaryBuilderBase? Without this, I’m not able to build on macOS 13..6.

PR #384, opened 2024-07-19
Add compat with HistoricalStdlibVersions v2

JamesWrigley:
This fixes support for nightly with the new Pkg version: https://github.com/JuliaLang/Pkg.jl/pull/3911. Would it be possible to make a new release once it's merged?

Example failure from the Clang.jl tests:
- https://github.com/JuliaInterop/Clang.jl/pull/502#issuecomment-2229472866
- https://github.com/JuliaInterop/Clang.jl/actions/runs/9964176318/job/27531889120?pr=502

Close #382.


JamesWrigley:
Hmm, this will need some more work to be backwards compatible.

JamesWrigley:
Alrighty, that should do the trick :crossed_fingers: Also closes #382.

JamesWrigley:
On second thoughts, I made HistoricalStdlibVersions v2 a requirement since we now require the new `StdlibInfo` data structure in it.

src/BinaryBuilderBase.jl
```diff
@@ -222,8 +222,18 @@ function __init__()
         use_ccache[] = true
     end
 
-    # Populate `Pkg.Types.STDLIBS_BY_VERSION`
-    append!(empty!(Pkg.Types.STDLIBS_BY_VERSION), HistoricalStdlibVersions.STDLIBS_BY_VERSION)
+    # Populate `Pkg.Types.STDLIBS_BY_VERSION`. On 1.12+ the stdlib information
+    # is a StdlibInfo object, but on earlier versions it's a tuple so we have to
+    # reconstruct the old tuple data structure from the StdlibInfo data
+    # structures in HistoricalStdlibVersions.STDLIBS_BY_VERSION.
+    if VERSION > v"1.11"
```
giordano, line 229: I'm not sure this condition is correct. Can you elaborate on words what you mean?

JamesWrigley: Sure, in https://github.com/JuliaLang/Pkg.jl/pull/3911 the `DictStdLibs` definition was changed: https://github.com/JuliaLang/Pkg.jl/pull/3911/files#diff-8387f5e55b4815b8e65004b6d2e4039e1f5cb8525cc9c513c69bee31fec44890L3-R14

Its values are now the new `StdlibInfo` type instead of a `Tuple{String,Union{VersionNumber,Nothing}}`. The `STDLIBS_BY_VERSION` variable is defined using `DictStdLibs`: https://github.com/JuliaLang/Pkg.jl/blob/e6880bc9d8a04d95df6e341c76786219a4efc33f/src/HistoricalStdlibs.jl#L22

In BinaryBuilderBase we populate `STDLIBS_BY_VERSION` using the values from `HistoricalStdlibVersions.STDLIBS_BY_VERSION`, which changed in https://github.com/JuliaPackaging/HistoricalStdlibVersions.jl/pull/23 to use the new `StdlibInfo`.

So when BinaryBuilderBase is loaded on nightly using the new Pkg version (which was added to nightly in https://github.com/JuliaLang/julia/pull/55112) but with the old compat of HistoricalStdlibs v1, it'll try to insert a dictionary of tuple values from HistoricalStdlibs into the dictionary in Pkg of `StdlibInfo` values, and fail. This change will make it only insert them directly on Julia 1.12+ with the new Pkg version, and otherwise fall back to recreating the tuple form and inserting those.

JamesWrigley: Oh, actually I'm an idiot :facepalm: I did not notice that v2 can actually do the whole thing for us: https://github.com/JuliaPackaging/HistoricalStdlibVersions.jl/blob/5879c5f690795208481c60b904f4af4e8c1eeef8/src/HistoricalStdlibVersions.jl#L38

Fixed in 966be26.

giordano: My point was that if you want to say "from v1.12 onwards" then `> v"1.11"` was wrong, because that matches already `v"1.11.1"`

JamesWrigley: Ah, yes you're quite right.

giordano:
> 

> Would it be possible to make a new release once it's merged?

I these cases bumping the version number of the package makes life easier for everybody and increases the probability of a new release happening.

JamesWrigley:
Bumped in 63702e3.

PR #385, opened 2024-07-28
improve error message for unsafe flags

SimonDanisch:
This makes it clearer, that Binary builder actually doesn't allow unsafe flags and errors.

### DIFF
--- a/src/Runner.jl
+++ b/src/Runner.jl
@@ -308,7 +308,7 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
         if length(unsafe_flags) >= 1
             write(io, """
             if [[ "\${ARGS[@]}" =~ \"$(join(unsafe_flags, "\"|\""))\" ]]; then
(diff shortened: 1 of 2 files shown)


src/Runner.jl
```diff
@@ -308,7 +308,7 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
         if length(unsafe_flags) >= 1
             write(io, """
             if [[ "\${ARGS[@]}" =~ \"$(join(unsafe_flags, "\"|\""))\" ]]; then
-                echo -e \"BinaryBuilder: You used one or more of the unsafe flags: $(join(unsafe_flags, ", "))\\nPlease repent.\" >&2
+                echo -e \"BinaryBuilder: You used one or more of the unsafe flags: $(join(unsafe_flags, ", "))\\nThis is not allowed, please remove all unsafe flags from your build script to continue.\" >&2
```
giordano, line 311: Something that I always wanted to fix: make the error message have `error:`, that makes finding the error in the log much easier:
```suggestion
                echo -e \"BinaryBuilder error: You used one or more of the unsafe flags: $(join(unsafe_flags, ", "))\\nThis is not allowed, please remove all unsafe flags from your build script to continue.\" >&2
```

PR #386, opened 2024-07-29
Add bootstrap artifacts for LLVM 14, 15 and 17.

maleadt:
Counterpart of https://github.com/JuliaPackaging/Yggdrasil/pull/9162.

@giordano Anything else I need to do? FWIW, I had to do the `Artifacts.toml` changes manually, for some reason `--deploy` didn't pick up my `dev`ed BinaryBuilderBase.jl.


giordano (approved):
> Anything else I need to do?

If we don't need to touch anything in `src/Runner.jl` or similar to adjust the compiler wrappers then that's it, yes, thanks!

> for some reason `--deploy` didn't pick up my `dev`ed BinaryBuilderBase.jl.

Uhm, that's weird, this is where the script picks the `Artifacts.toml` file: https://github.com/JuliaPackaging/Yggdrasil/blob/08aaaaf1af11396ae31ecf39d7e1501356f02c66/0_RootFS/common.jl#L61  It relies on `pathof(BinaryBuilderBase)` giving the right path

PR #387, opened 2024-08-12
Add support for FreeBSD AArch64

ararslan:
Accompanying Yggdrasil PR: https://github.com/JuliaPackaging/Yggdrasil/pull/9220

I'll flip this out of draft mode once all compilers have been built.

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -482,6 +482,160 @@ os = "linux"
     sha256 = "3599936852e8295ced08fa92eac970cfae73c0e686f21e1c54ec2551157672af"
     url = "https://github.com/JuliaPackaging/Yggdrasil/releases/download/GCCBootstrap-v9.1.0+2/GCCBootstrap-aarch64-linux-musl.v9.1.0.x86_64-linux-musl.unpacked.tar.gz"
 
+[["GCCBootstrap-aarch64-unknown-freebsd13.2.v10.2.0.x86_64-linux-musl.squashfs"]]
+arch = "x86_64"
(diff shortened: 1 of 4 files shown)


giordano:
With reference to https://github.com/JuliaPackaging/Yggdrasil/pull/9220#issuecomment-2284551377, for gcc I was thinking of https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/56580d9a65804071e2d3626b82283f5978ac8d14/src/Rootfs.jl#L578-L584 but that was specific to aarch64-darwin where we have a different toolchain version number under the hood (the user doesn't need to know it), but this may just work for aarch64-freebsd https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/56580d9a65804071e2d3626b82283f5978ac8d14/src/Rootfs.jl#L586-L593

However, if we only have GCC >= 7, you probably need to adapt https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/56580d9a65804071e2d3626b82283f5978ac8d14/src/Rootfs.jl#L763-L770 gcc 7 has libgfortran v4, so you have to exclude v3

giordano:
Reminder: you'll need to add aarch64-freebsd to the list of platforms incompatible with Julia v1.5 at https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/8aadff4c54e12a31fe605defd81d99da78c60af2/src/AutoBuild.jl#L199

ararslan:
Ah thank you! PR: https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1342

ararslan:
I also bumped the version so that we can tag a new release.

giordano (approved):

Issue #347, opened 2023-11-04
[CI] Sandbox error when running shards testuite on GitHub Actions


giordano:
> I finally managed to get some more useful error messages, for the failure at https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/8d1eca1e3a63a2de30d9a9eee7381c9ca999ea60/test/runners.jl#L446
> ```
> rustc -o /tmp/testsuite/x86_64-linux-gnu/rust/hello_world/hello_world hello_world.rs
>   Installing cargo_build v0.1.0 (/usr/share/testsuite/rust/cargo_build)
> error: failed to compile `cargo_build v0.1.0 (/usr/share/testsuite/rust/cargo_build)`, intermediate artifacts can be found at `/usr/share/testsuite/rust/cargo_build/target`.
> To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
> 
> Caused by:
>   Cross-device link (os error 18)
> make[1]: *** [Makefile:7: /tmp/testsuite/x86_64-linux-gnu/rust/cargo_build/cargo_build] Error 101
> make: *** [Makefile:48: install-project-rust-cargo_build] Error 2
> ```
> Sounds like a problem in the sandbox code, which perhaps only arises on the CI setup.

_Originally posted by @giordano in https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/346#issuecomment-1793431598_

Note: this happens only with the non-squashfs images with UserNS runners (not Docker).

PR #390, opened 2024-09-25
Add bootstrap artifacts for LLVM 18.

maleadt:
x-ref https://github.com/JuliaPackaging/Yggdrasil/pull/9415

cc @ararslan @giordano


ararslan:
The CI failure is because a `@test_broken` passed. Perhaps this PR magically fixes https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/347 for some completely unknown reason?

ararslan:
Also, do you want to bump the version in Project.toml for a new release as part of this PR? If not, someone can just do it in a follow-up PR.

giordano:
Sounds unlikely it's related to this PR, especially because it's in the rust toolchain, maybe some underlying change in github actions environment. I presume running ci on a separate pr would show the same

PR #394, opened 2024-12-14
Add Bazel toolchain

mofeing:
cc @giordano @gbaraldi

@wsmoses given that Bazel makes it so difficult to get values from env vars, this is actually easier

### DIFF
--- a/src/BuildToolchains.jl
+++ b/src/BuildToolchains.jl
@@ -2,6 +2,7 @@ abstract type AbstractBuildToolchain{C} end
 
 struct CMake{C} <: AbstractBuildToolchain{C} end
 struct Meson{C} <: AbstractBuildToolchain{C} end
+struct Bazel{C} <: AbstractBuildToolchain{C} end
 
 c_compiler(::AbstractBuildToolchain{:clang}) = "clang"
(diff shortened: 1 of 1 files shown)


src/BuildToolchains.jl
```diff
@@ -299,3 +310,168 @@ function cargo_config_file!(dir::AbstractString, platform::AbstractPlatform;
...
+                # "/opt/$(aatarget)/$(aatarget)/include/c++/10.2.0",
+                # "/opt/$(aatarget)/$(aatarget)/include/c++/10.2.0/$(aatarget)",
+                # "/opt/$(aatarget)/$(aatarget)/include/c++/10.2.0/backward",
+                # "/opt/$(aatarget)/$(aatarget)/include/c++/10.2.0/parallel",
+                "%sysroot%/../include",
+                "%sysroot%/usr/include",
+                "%sysroot%/../include/c++/10.2.0",
+                "%sysroot%/../include/c++/10.2.0/$(aatarget)",
+                "%sysroot%/../include/c++/10.2.0/backward",
+                "%sysroot%/../include/c++/10.2.0/parallel",
```
giordano, line 367: Hardcoding paths like this doesn't seem good, also because they're specific to the GCC version used.

PR #399, opened 2024-12-26
Allow specifying a Rust and Go toolchain version for the build

imciner2:
This exposes two new keyword arguments for finding the shards: `preferred_rust_version` and `preferred_go_version`, which will specify the version of the toolchain to use. Note these are implemented as strict requests for now, but I kept the "preferred" name to match the gcc and llvm arguments.

I know the Rust toolchain is probably fine as a strict request, but I don't know about Go's commitment to compatibility across compiler versions.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "BinaryBuilderBase"
 uuid = "7f725544-6523-48cd-82d1-3fa08ff4056e"
 authors = ["Elliot Saba <staticfloat@gmail.com>"]
-version = "1.33.0"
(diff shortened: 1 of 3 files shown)


giordano:
Can you please bump the minor version of this package so that we can tag it and then require that version in https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1358 for consistency?  Looks good to me otherwise, this is pretty much what I was expecting (well, even more than I was thinking).

imciner2:
Version bumped.

giordano (approved):

PR #407, opened 2025-01-30
[Runner] Work around wrong path of compiler libraries for riscv64. Again

giordano:
I don't know how this has happened, because I thought this was resolved by [JuliaPackaging/Yggdrasil@`123ac9b` (#10132)](https://github.com/JuliaPackaging/Yggdrasil/pull/10132/commits/123ac9b8f3f59936c63c5772dec5d8541de8596d) (and the path is right for x86_64, aarch64, and powerpc64le), but the compiler support libraries ended up again in `lib/` instead of `lib64/` for riscv64, so we have to repeat #402.  I thought we had a test about this (create a C++ library, and link to it with `cc`), apparently we don't, probably we should, but I don't have the time to add it right now.


giordano:
Ok, I believe this should be working now, tests were failing before because I was testing `clang` as well: we're passing `-L/opt/${target}/${target}/${lib64} -Wl,-rpath-link,/opt/${target}/${target}/${lib64}` only to `gcc`, we may want to do it for clang too, but leaving that change for the future for the time being.

PR #415, opened 2025-02-15
[Dependencies] Do not pass the content to `Pkg.add` in recent Julia

giordano:
CC @IanButterworth.


src/Prefix.jl
```diff
@@ -615,6 +615,15 @@ function Pkg_add(args...; kwargs...)
     end
 end
 
+function Pkg_add(ctx::Pkg.Types.Context, rest...; kwargs...)
+    @static if VERSION >= v"1.13-"
+        # No need to pass the context anymore.
+        _Pkg_add(rest...; kwargs...)
```
IanButterworth, line 621: This needs fixing in Pkg, but as a workaround this might work?
```suggestion
        _Pkg_add(rest...; ctx.julia_version, kwargs...)
```

giordano: I can add it if still need, but it doesn't look like it helps on `master` 😞

PR #418, opened 2025-02-18
Update FreeBSD versions to 13.4 on x86-64 and 14.1 on AArch64

ararslan:
Companion to https://github.com/JuliaPackaging/Yggdrasil/pull/10402

To do:
- [x] Update PlatformSupport
    - [x] Fix the ones renamed from previous releases so that the tarballs unpack with an appropriately named directory
- [x] GCC
    - [x] All versions for FreeBSD 13.4 x86-64
    - [x] Versions 9+ for FreeBSD 14.1 AArch64
    - [x] Either fix 7 and 8 for FreeBSD 14.1 AArch64 for parity with current 13.2 or update the package code to use 9+
- [x] Remove artifacts for FreeBSD 13.2 (except older PlatformSupport) since those always seem to interfere (unless we do #419 and it actually works)
- [x] Bump package version for a new release

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -3914,6 +3914,28 @@ os = "linux"
     sha256 = "ade3abf79b76b3ec94e078246fbd78e935bd51c6f69b894fa91c0cd580db6045"
     url = "https://github.com/JuliaPackaging/Yggdrasil/releases/download/PlatformSupport-v2024.8.10/PlatformSupport-aarch64-apple-darwin20.v2024.8.10.x86_64-linux-musl.unpacked.tar.gz"
 
+[["PlatformSupport-aarch64-apple-darwin20.v2025.2.15.x86_64-linux-musl.squashfs"]]
+arch = "x86_64"
+git-tree-sha1 = "6f92bdb4b289a54e04c9b812eb74f80cd8228326"
+lazy = true
+libc = "musl"
+os = "linux"
+
+    [["PlatformSupport-aarch64-apple-darwin20.v2025.2.15.x86_64-linux-musl.squashfs".download]]
+    sha256 = "66493431b2d377cdd8811c5bf1e950c7fe07f81ec476ca440de1534a91f580b7"
+    url = "https://github.com/JuliaPackaging/Yggdrasil/releases/download/PlatformSupport-v2025.2.15/PlatformSupport-aarch64-apple-darwin20.v2025.2.15.x86_64-linux-musl.squashfs.tar.gz"
+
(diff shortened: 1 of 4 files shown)


ararslan:
There's one decision remaining here, and it's in regard to how we want to handle GCC.

All GCC versions built successfully for FreeBSD 13.4 on x86-64, but only 9+ built for FreeBSD 14.1 on AArch64; GCC 7 and 8 failed with a linker error. For those versions, it's complaining about an unrecognized file format for `crtn.o` in the sysroot, but I haven't found anything obviously amiss with that file. The `file` command recognizes it as a FreeBSD ELF file, and `readelf` can read it. My knowledge of object files is admittedly rather limited, and Google hasn't been much help regarding the error, so I'm not sure where I'd go from here for further debugging. My only guess would be that the system compiler on 14.1 (Clang 18, IIRC) is too new to build old GCC, but I'm not confident about that assessment nor would I know how to address it.

What I've implemented as a stopgap is to remove the artifacts for GCC 9+ on 13.2 AArch64 but keep the existing artifacts for 7 and 8. This means that a build preferring an earlier GCC version will actually be targeting 13.2 instead of 14.1 for any component compiled with the GCC toolchain, so OS version targets may be mixed. That shouldn't be an issue in theory, and indeed, I've been using binaries built for 13.2 on 14.1 _mostly_ successfully. However, I did encounter an issue with MPFR that was fixed by using a native build, which suggests that the problem is with the artifact built for 13.2. Newer builds would be targeting 14.1... that is, unless they need GCC 7 or 8, I suppose.

So I guess the question is this: Which is better, what I've implemented currently and described above, or dropping GCC 7 and 8 and requiring 9+ on 14.1 AArch64? I don't have a good sense of the implications of the latter, including any potential for breakage.

giordano:
I think dropping GCC 7 and 8 is fine.  Note you'll have to restrict expand_libgfortran to v5 for this platform (as we do already for aarch64-darwin and riscv64)

ararslan:
Is this good to go, @giordano? Also, can you remind me whether we need to update a manifest somewhere once a new release here is registered or if the latest version gets used automatically?

ararslan:
Can you also remind me of the proper merge order of this and https://github.com/JuliaPackaging/Yggdrasil/pull/10402?

giordano (approved):
> can you remind me whether we need to update a manifest somewhere once a new release here is registered or if the latest version gets used automatically?

https://github.com/JuliaPackaging/Yggdrasil/actions/workflows/update_manifest.yml

> Can you also remind me of the proper merge order of this and https://github.com/JuliaPackaging/Yggdrasil/pull/10402?

No particular order

PR #420, opened 2025-04-07
RISCV: Add CMake variable to set the system library directory

imciner2:
RISCV installs place the libraries inside a folder that is named after the ABI, so we need to prompt CMake to search there. This should fix the toolchain issue discussed in https://github.com/JuliaPackaging/Yggdrasil/pull/10922#discussion_r2029505751.

cc @eschnett @giordano

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "BinaryBuilderBase"
 uuid = "7f725544-6523-48cd-82d1-3fa08ff4056e"
 authors = ["Elliot Saba <staticfloat@gmail.com>"]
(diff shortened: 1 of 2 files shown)


imciner2:
I guess I should probably also do a version number bump in this? (it seems like a patch release type of change to me).

giordano (approved):
Minor version bump is fine, but in Yggdrasil with the [Update Manifest](https://github.com/JuliaPackaging/Yggdrasil/actions/workflows/update_manifest.yml) action we just checkout the tip of the default branch, don't need necessarily a new version tag.

PR #425, opened 2025-05-08
Implement preliminary OCaml support

topolarity:
This is a preliminary implementation of OCaml compiler support in BinaryBuilder.

Dependent on https://github.com/JuliaPackaging/Yggdrasil/pull/11146

### DIFF
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -383,6 +383,10 @@ struct RustBuild <: CompilerBuild
     version::VersionNumber
 end
 
+struct OCamlBuild <: CompilerBuild
+    version::VersionNumber
+end
+
 getversion(c::CompilerBuild) = c.version
 getabi(c::CompilerBuild) = c.abi
 
@@ -417,6 +421,8 @@ const available_llvm_builds = LLVMBuild.(get_available_builds("LLVMBootstrap."))
 const available_go_builds = GoBuild.(get_available_builds("Go."))
 
 const available_rust_builds = RustBuild.(get_available_builds("RustBase."))
+
+const available_ocaml_builds = OCamlBuild.(get_available_builds("OCaml-"))
 
 """
     gcc_version(p::AbstractPlatform, GCC_builds::Vector{GCCBuild},
@@ -586,6 +592,7 @@ function choose_shards(p::AbstractPlatform;
             LLVM_builds::Vector{LLVMBuild}=available_llvm_builds,
             Rust_builds::Vector{RustBuild}=available_rust_builds,
             Go_builds::Vector{GoBuild}=available_go_builds,
+            OCaml_builds::Vector{OCamlBuild}=available_ocaml_builds,
             archive_type::Symbol = (use_squashfs[] ? :squashfs : :unpacked),
             bootstrap_list::Vector{Symbol} = bootstrap_list,
             # Because GCC has lots of compatibility issues, we always default to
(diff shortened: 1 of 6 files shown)


src/Runner.jl
```diff
@@ -631,6 +631,14 @@ function generate_compiler_wrappers!(platform::AbstractPlatform; bin_path::Abstr
     end
     gofmt(io::IO, p::AbstractPlatform) = wrapper(io, "/opt/$(host_target)/go/bin/gofmt"; allow_ccache=false)
 
+    # OCaml stuff
+    function ocaml_wrapper(io::IO, tool::String, p::AbstractPlatform)
+        return wrapper(io, "/opt/$(aatriplet(p))/bin/$(aatriplet(p))-$(tool)")
```
topolarity, line 636: This path is wrong, but I wasn't quite sure where the RootFS artifacts would end up

topolarity:
This looks to be almost working. There is some slight disagreement about `darwinX` in the target tuples:
```console
sandbox:${WORKSPACE}/srcdir # /opt/aarch64-apple-darwin/bin/ocamlopt.opt
No input files
sandbox:${WORKSPACE}/srcdir # ocamlopt.opt
/opt/bin/aarch64-apple-darwin20-libgfortran5-cxx11/ocamlopt.opt: line 5: /opt/aarch64-apple-darwin20/bin/ocamlopt.opt: No such file or directory
```

Any preference for how to resolve that?

maleadt:
Okay, this should be the first in the OCaml series (also including https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/1382 and https://github.com/JuliaPackaging/Yggdrasil/pull/11146) that's ready for review, merge and tag. It's all fairly straightforward, only involving OCaml, with the exception of two broadly-impacting changes:
- `preferred_gcc_version` now defaults to `nothing`, indicating unspecifiedness, which I use to bump it from the default `4.8` to `6` when OCaml is requested. This because the OCaml compiler shards themselves are built using GCC v6.
- symlinks to wrappers are now also provided for host platform binaries, making it possible to switch to the host toolchain by doing `PATH=/opt/bin/$host_full:/opt/$host/bin:$PATH`. Before, there was no easy way to e.g. get a `as` that mapped to `x86_64-linux-musl-as`, which was problematic for my `ocamlopt`-based recipe that simply executes `as`.

src/Rootfs.jl
```diff
@@ -599,8 +606,21 @@ function choose_shards(p::AbstractPlatform;
             preferred_rust_version::VersionNumber = maximum(getversion.(Rust_builds)),
             # Always default to the latest Go version
             preferred_go_version::VersionNumber = maximum(getversion.(Go_builds)),
+            # Always default to the latest OCaml version
+            preferred_ocaml_version::VersionNumber = maximum(getversion.(OCaml_builds)),
         )
 
+    # The preferred GCC version depends on the compilers we are using.
+    if preferred_gcc_version === nothing
+        preferred_gcc_version = if :ocaml in compilers
```
giordano, line 615: So, I think this is ok, but I wanted to point out that we already do some shenanigans in `gcc_version` at https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/dd705399548bb909f4c6e5f815b6de917f746faa/src/Rootfs.jl#L447-L538 for example in certain cases we change the version of the gcc toolchain based on the...llvm toolchain.  I was never proud of that mess though.

giordano (approved):

PR #433, opened 2025-07-15
Add tests to ensure tarballs are compressed with the expected format

giordano:


test/archive_utils.jl
```diff
@@ -27,14 +27,17 @@ using Test
 
             mktempdir() do output_dir
                 for (format, ext, hash) in (("gzip", "gz", "568f743e965b63d3187b6a2647700a71d1d7520b4596fbf2bfb39ffa67c4bb55"),
-                                            # Compressing with p7zip/xz doesn't seem to be fully reproducible, at least not
-                                            # across different systems. We'll try to investigate more, but for the time being
-                                            # skip the reproducibility test for it.
-                                            ("xz", "xz", ""))
+                                            ("xz", "xz", "293736704ee772836edf67ba4bac55c9604721d6df332cef9bdf4c9c06b39a8c"))
```
giordano, line 30: This is now reproducible with `xz` instead of `p7zip` 🥳

giordano (commented):
.

PR #435, opened 2025-07-29
Add creation of the build-id to the linker flags

imciner2:
This is the toolchain part of the implementation of the build-id. This should now make the linker build in a build-id to the libraries/executables, and use one that also allows reproducibility.

ref: https://github.com/JuliaPackaging/Yggdrasil/issues/11013, https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/1272

### DIFF
--- a/test/runners.jl
+++ b/test/runners.jl
@@ -2,6 +2,7 @@ using Test
 using BinaryBuilderBase
 using BinaryBuilderBase: platform_dlext, platform_exeext, prefer_clang
 using Pkg
+using ObjectFile
 
 @testset "Wrappers utilities" begin
     @test nbits(Platform("i686", "linux")) == 32
@@ -171,21 +172,26 @@ end
                 """
             test_script = """
                 set -e
+                cd /workspace
                 # Make sure setting `CCACHE` doesn't affect the compiler wrappers.
                 export CCACHE=pwned
                 export USE_CCACHE=false
                 echo '$(test_c)' > test.c
                 # Build shared library
                 $(compiler) -shared test.c -o libtest.\${dlext}
-
-                # Print out the notes in the library
-                readelf -n libtest.\${dlext}
                 """
             cmd = `/bin/bash -c "$(test_script)"`
             @test run(ur, cmd, iobuff)
-            seekstart(iobuff)
-            # Make sure the compiled library has the note section for the build-id
-            @test occursin("NT_GNU_BUILD_ID", readchomp(iobuff))
+
+            # Load the library file and test it for the build-id
+            lib_path = joinpath(dir, "libtest."*platform_dlext(platform))
+            lib = open(lib_path)
+            obj_handles = readmeta(lib)
+            obj = first(obj_handles)
+            secs = Sections(obj)
+
+            # The section must exist for the build-id to be present
+            @test !isnothing(findfirst(s -> section_name(s) == ".note.gnu.build-id", secs))
         end
     end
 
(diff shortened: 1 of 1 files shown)


imciner2:
Forgot that build-id is an ELF only thing, so we can only add it on Linux and FreeBSD.

giordano:
Can we have some tests?

imciner2:
I've blind-coded a test for this based on what I think the output should be. Let's see if that passes. it's a bit awkward because we don't have the actual elf to query here, so I had to do a `readelf` inside the script and just parse the output.

test/runners.jl
```diff
@@ -153,6 +155,40 @@ end
...
+                $(compiler) -Werror -shared test.c -o libtest.\${dlext}
+
+                # Print out the notes in the library
+                readelf -n libtest.\${dlext}
+                """
+            cmd = `/bin/bash -c "$(test_script)"`
+            @test run(ur, cmd, iobuff)
+            seekstart(iobuff)
+            # Make sure the compiled library has the note section for the build-id
+            @test occursin(r".note.gnu.build-id", readchomp(iobuff))
```
giordano, line 188: Probably would work without a regex (otherwise we'd want to escape the dots)?
```suggestion
            @test occursin(".note.gnu.build-id", readchomp(iobuff))
```

giordano:
Can ObjectFile.jl read the notes?

imciner2:
> Can ObjectFile.jl read the notes?

It can, but I didn't see any place where we can get the actual compiled file out in these tests already (or at least I didn't see any tests doing it), and we didn't have ObjectFile.jl in the deps already, so I didn't dive deeper into trying to make that work.

Using ObjectFile.jl, this is pseudo-code to test if the build-id section exists (and can also form the basis for an Audit pass in BinaryBuilder.jl):
```
l = open(<lib>)
obj_handles = readmeta(l)
obj = first(obj_handles)
sec = ELFSections(obj)

# This will be true if there is no section
findfirst(s -> section_name(s) == ".note.gnu.build-id", sec) === nothing
```

giordano:
This broke reproducibility of Windows binaries with GCC 5+: https://github.com/JuliaPackaging/BinaryBuilder.jl/blob/d41ddc84157a8c4b490761b0fb5f66cf32746ce6/test/auditing.jl#L796-L839

vtjnash:
On Windows I'm told that this should appear to be a copy of the CodeView GUID value already in the binary

giordano:
Reproducer:
```console
julia> BinaryBuilderBase.runshell(Platform("i686", "windows"); preferred_gcc_version=v"6");
sandbox:${WORKSPACE} # echo 'int foo(){ return 42; }' | SUPER_VERBOSE=1 cc -x c -shared - -o libfoo.${dlext} -Wl,--out-implib,libfoo.${dlext}.a
ccache /opt/i686-w64-mingw32/bin/i686-w64-mingw32-gcc -D_GLIBCXX_USE_CXX11_ABI=1 -frandom-seed=0x8ac3fab9 -DWINVER=0x0A00 -D_WIN32_WINNT=0x0A00 -march=pentium4 -mtune=generic -x c -shared - -o libfoo.dll -Wl,--out-implib,libfoo.dll.a -Wl,--no-insert-timestamp -Wl,--build-id=sha1
sandbox:${WORKSPACE} # sha256sum libfoo.dll*
edfd3305aec90e38954c80ef2d3fea1d9a966abf16602a0d157c7cff5c7cfbc1  libfoo.dll
c20ac05a44844b9d3180ac2ba1e206e9a11e247bc6617f4a58cd5ee216991d5c  libfoo.dll.a
sandbox:${WORKSPACE} # echo 'int foo(){ return 42; }' | SUPER_VERBOSE=1 cc -x c -shared - -o libfoo.${dlext} -Wl,--out-implib,libfoo.${dlext}.a
ccache /opt/i686-w64-mingw32/bin/i686-w64-mingw32-gcc -D_GLIBCXX_USE_CXX11_ABI=1 -frandom-seed=0x8ac3fab9 -DWINVER=0x0A00 -D_WIN32_WINNT=0x0A00 -march=pentium4 -mtune=generic -x c -shared - -o libfoo.dll -Wl,--out-implib,libfoo.dll.a -Wl,--no-insert-timestamp -Wl,--build-id=sha1
sandbox:${WORKSPACE} # sha256sum libfoo.dll*
19a31cc606dc96cf7aafb8b2a6929a6cd810c9584f7936a36410266c1e4cac14  libfoo.dll
7eb234fd60bd6d17ef6b5753819d6aac7c5a7762ba690d74284eeffab006e145  libfoo.dll.a
```
I tried to compare the two dlls with https://try.diffoscope.org, but it crashed with an OOM.  However this may be specific to the i686 toolchain, I believe the x86_64 one is still reproducible.

imciner2:
> On Windows I'm told that this should appear to be a copy of the CodeView GUID value already in the binary

Do you know if that is implemented in `ld` for Windows?

Looking at the current source tree for binutils (https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=ld/ldbuildid.c;h=a4a6dc3c3e9403e72bcd52eaa6dfda39fd912586;hb=HEAD), it seems our options for build-id types are: `md5`, `sha1`, `xx` (XXHash), UUID, and self-generated hex. The UUID that is used is generated by the [`UuidCreate`](https://learn.microsoft.com/en-us/windows/win32/api/rpcdce/nf-rpcdce-uuidcreate) in the RPC library.

Right now, we have it set to SHA1, because that was supposed to be reproducible.

PR #444, opened 2025-09-27
Bump patch version to v1.41.1

Zentrik:
I think we need a new release before the fixes for https://github.com/JuliaPackaging/Yggdrasil/pull/12004 will take effect.


giordano:
Releasing a new version is fine, but in Yggdrasil we check out commits on `master` of this repo, and we upgraded it a couple of weeks ago: https://github.com/JuliaPackaging/Yggdrasil/pull/12075: #443 was needed also for https://github.com/JuliaPackaging/Yggdrasil/pull/12073.

PR #446, opened 2025-10-19
[Rootfs] Add keyword argument to exclude old ABIs from output of `expand_*` functions

giordano:
This was [announced](https://discourse.julialang.org/t/psa-intent-to-deprecate-cxx03-libgfortran3-libgfortran4-abis/130799) a few months ago and no one reacted, this implements the suggested changes.

My only question is: do we want to retain the ability for these functions to include the old ABIs with some extra keyword argument, or we simply forget about them (users will still be able to manually create those platforms)?  This PR does the latter.  If we want the former, what's the suggested keyword argument?

### DIFF
--- a/src/Rootfs.jl
+++ b/src/Rootfs.jl
@@ -831,16 +831,18 @@ function supported_platforms(;exclude::Union{Vector{<:Platform},Function}=Return
 end
 
 """
-    expand_gfortran_versions(p::AbstractPlatform)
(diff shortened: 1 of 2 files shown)


Keno:
Keeping the old ABIs with a flag, seems like a good idea to me, just in case we need to rebuild something old for a security fix.

staticfloat:
I agree, we can just add a kwarg like `; old_platforms = false` or something.

giordano:
@Keno @staticfloat are you happy with this?

PR #451, opened 2025-11-04
Replace unzip_jll with p7zip_jll

nhz2:
Fix #450


nhz2:
It turns out that p7zip_jll has a bug on Linux https://github.com/JuliaLang/julia/issues/60071

Maybe we need to wait until the fix in https://github.com/JuliaLang/julia/pull/60025 is backported

ViralBShah:
Is this mergeable now?

ViralBShah:
@giordano @eschnett Hopefully this is a simple one to merge.

PR #470, opened 2026-05-06
Fetch all tags in LibGit2.fetch

odow:
See the discussion in https://github.com/JuliaPackaging/Yggdrasil/pull/13647/changes

I'm not sure if this is correct (a bot helped me). Or how to test it, because it requires the cached repo and mucking about upstream.

### DIFF
--- a/src/Sources.jl
+++ b/src/Sources.jl
@@ -218,7 +218,13 @@ function cached_git_clone(url::String;
             # this is not only faster, it avoids race conditions when we have
             # multiple builders on the same machine all fetching at once.
             if hash_to_check === nothing || !LibGit2.iscommit(hash_to_check, repo)
-                LibGit2.fetch(repo)
+                LibGit2.fetch(
+                    repo;
+                    refspecs = [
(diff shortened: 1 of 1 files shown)


giordano:
I _think_ the nicer way to do it is to set the download tags fetch option to "all", only problem being that julia currently doesn't expose it in `LibGit2.fetch` as an option: https://github.com/JuliaLang/julia/blob/328b7d27fd538e6deff6602823fff1b2f393c27e/stdlib/LibGit2/src/LibGit2.jl#L276-L318. It should be added as an argument and passed to the fetch option constructor.

odow:
I've at least verified that the ref spec are correct:
```
(base) odow@Mac clp-git % git fetch origin '+refs/heads/*:refs/remotes/origin/*'
(base) odow@Mac clp-git % git fetch origin '+refs/tags/*:refs/tags/*'  
```
whereas making a typo results in
```
(base) odow@Mac clp-git % git fetch origin '+refs/tags/*:refs/tags/' 
fatal: invalid refspec '+refs/tags/*:refs/tags/'
```

PR #477, opened 2026-06-19
WIP: Add UCRT rootfs support

Keno:
As we found out in https://github.com/JuliaLang/julia/pull/62152#issuecomment-4747173255, mingw13 breaks its ABI. Since moving to UCRT is long overdue anyway, this repurposes the recently updated mingw13 rootfs shards as -ucrt-minw32 shards. In theory it is possible to build non-ucrt builds on ming13 and vice versa, but if we're not doing the ABI break, then I think it's cleaner to just say everything mingw13 and up is ucrt and everything without the ABI break is msvcrt. Hopefully we'll be able to transition to ucrt quickly enough that we don't need any mingw fixes to the old ABI (although that would be possible of course).

In some sense an alternative to https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/476, but that can be merged first of course.


Keno:
Actually, I realized that libstdc++ also links libc, so it's not actually that simple and we do need to separate if only for the runtime libraries.

PR #489, opened 2026-08-21
Use dedicated gzip and zstd decompressors

maleadt:
Use `pigz_jll` for gzip decompression and `Zstd_jll` for zstd streams instead of routing both through p7zip.

The command-returning JLL wrappers carry their environment in the `Cmd`, avoiding the process-global environment mutation performed by the do-block wrappers. This makes `unpack` safe to call concurrently.

Using pigz also improves gzip decompression throughput. On 18 Clang_unified tarballs, unpacking decreased from 26.7 seconds to 18.8 seconds.

This work prompted by an excruciatingly slow registration of Clang_unified_jll on Yggdrasil.

Assisted by: Claude Fable 5, GPT 5.6 Sol

### DIFF
--- a/src/ArchiveUtils.jl
+++ b/src/ArchiveUtils.jl
@@ -1,5 +1,5 @@
 using Base: SHA1
-using Downloads, Tar, p7zip_jll, pigz_jll, SimpleBufferStream, SHA
+using Downloads, Tar, p7zip_jll, pigz_jll, SimpleBufferStream, SHA, Zstd_jll
 using Pkg.Artifacts: artifact_exists, artifact_path, query_override
 using XZ_jll: xz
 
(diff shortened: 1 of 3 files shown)


src/ArchiveUtils.jl
```diff
@@ -21,6 +21,27 @@ function detect_compressor(header::Vector)
     return nothing
 end
 
+"""
+    decompress_cmd(path, compressor) -> Cmd
+
+Build the command that decompresses `path` (already classified as `compressor`) to stdout.
+
+Note that we deliberately use the plain, non-`do`-block forms of the JLL executable
```
giordano: This paragraph sounds like too much information

giordano (approved):
Looks good, besides the excessive docstrings.

maleadt:
Yeah sorry, the initial push was only meant for CI, but I forgot to mark it draft.
Switched to Zstd_jll for 1.7 compatibility, and shortened comments/commits.

Issue #395, opened 2024-12-21
Rust toolchain tests failing with unpriviliged runner and unpacked compiler shards


giordano:
With
```
BINARYBUILDER_RUNNER=unprivileged
BINARYBUILDER_USE_SQUASHFS=false
```
Rust toolchain tests are [failing on CI](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/actions/runs/12446391255/job/34748724430#step:7:219) with
```
  Installing cargo_build v0.1.0 (/usr/share/testsuite/rust/cargo_build)
error: failed to compile `cargo_build v0.1.0 (/usr/share/testsuite/rust/cargo_build)`, intermediate artifacts can be found at `/usr/share/testsuite/rust/cargo_build/target`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  failed to create directory `/usr/share/testsuite/rust/cargo_build/target`

Caused by:
  Cross-device link (os error 18)
make[1]: *** [Makefile:7: /tmp/testsuite/x86_64-linux-gnu/rust/cargo_build/cargo_build] Error 101
```
This look like https://github.com/rust-lang/rustup/issues/2949.

giordano:
Presumably, setting [`CARGO_TARGET_DIR`](https://doc.rust-lang.org/cargo/reference/environment-variables.html) to an actually writable directory, as suggested by the error message, could help
> `CARGO_TARGET_DIR` — Location of where to place all generated artifacts, relative to the current working directory. See [build.target-dir](https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir) to set via config.

PR #345, opened 2023-11-02
Make lld more first party in BBB

gbaraldi:
I'm not really in love of how this is getting, I think a nice refactor might be to move all these gcc versions/platform to a toolchain struct that has everything in it. But for now this is an attempt

PR #490, opened 2026-08-29
Let Meson use lld with Clang when the compiler wrappers do

ChrisRackauckas-Claude:
> **Draft — please ignore until reviewed by @ChrisRackauckas.**

Follow-up to https://github.com/JuliaPackaging/Yggdrasil/pull/14605#discussion_r3887299826, where the Serd recipe had to `sed` the `*_ld` entries out of `${MESON_TARGET_TOOLCHAIN}` on macOS and @giordano asked for this to be fixed here instead.

### What changed and why

`toolchain_file(::Meson, ...)` hardcoded `clang_use_lld=false` (since #345), so the Clang cross file always named the cctools/binutils `ld` wrapper as `c_ld`/`cpp_ld`, even though the Clang wrapper itself links with lld by default whenever `clang_use_lld` is on (Clang ≥ 16 and GCC ≥ 6, i.e. `aarch64-apple-darwin` always and `x86_64-apple-darwin` with `preferred_gcc_version ≥ v"6"`). On macOS Meson cannot identify that linker: `-Wl,--version` gets `ld: unknown option: --version` from cctools, and Meson's Apple fallback probe `-Wl,-v` runs *without* the `-fuse-ld` override, so it hits the wrapper's default `ld64.lld`, whose banner has no `PROJECT:ld` line. Result: `meson.build:4:0: ERROR: Unable to detect linker for compiler ...`.

The fix honours `clang_use_lld` like the CMake toolchain already does, but passes the **flavour name** `lld` (the same way the GCC file passes `bfd`) instead of the path to the lld wrapper. The path variant does not work: with `-fuse-ld=<absolute path>` Clang does not know the linker is lld, so its Darwin driver emits legacy `-macosx_version_min` flags (and no `-platform_version`), which LLD 22 rejects with `ld64.lld: error: must specify -platform_version`. With `-fuse-ld=lld` Clang resolves `ld64.lld`/`ld.lld` from `PATH` (`/opt/bin/<target>/`) and links exactly as the plain `cc` wrapper does.

Behaviour is unchanged when `clang_use_lld` is `false` (the `-ld` wrapper path is still used, which Meson identifies through the cctools/binutils `-v`/`--version` output) and for the GCC cross files (`bfd`). It does change what Meson passes with Clang on Linux/FreeBSD/Windows when `clang_use_lld` is on: Meson now uses lld there too, matching the compiler wrapper; previously it forced the (often very old) binutils `ld` — e.g. Clang + Meson on `x86_64-linux-gnu` with the default GCC 4.8 shard fails on master with `x86_64-linux-gnu-ld: unrecognized relocation (0x2a)` (bfd 2.24), and builds with this PR when `clang_use_lld` is on.

Tests added:
- `Meson toolchain files`: checks the `c_ld`/`cpp_ld` entries of the generated cross files for both compilers and both `clang_use_lld` values.
- `Meson - <platform> - <compiler> - clang_use_lld=<bool>` in the runner tests: `meson setup` + `meson compile` + `meson install` of a one-file shared library with `${MESON_TARGET_TOOLCHAIN%.meson}_<compiler>.meson`. Covers the host platform by default and all of `supported_platforms()` under `BINARYBUILDER_FULL_SHARD_TEST`.

### Verification

All runs on Linux x86_64 with the unprivileged runner, BinaryBuilder 0.6.6, this package `Pkg.develop`ed, Meson 1.8.0 / Clang 22.1.7 / LLD 22.1.7 from the current rootfs and LLVMBootstrap shards.

**Failing before / passing after.** Content test on master:
```
Meson toolchain files: Test Failed at unit_test.jl:14
  Expression: occursin("\nc_ld = '$(linker)'\n", clang)
… (4 failures: aarch64/x86_64-apple-darwin, x86_64-linux, x86_64-freebsd, x86_64-windows with clang_use_lld=true)
```
with this PR: `Meson toolchain files |   40     40`.

Runner test (host + aarch64-macos + x86_64-macos + x86_64-windows + x86_64-linux + x86_64-freebsd, clang/gcc, clang_use_lld true/false) on master fails only in the cell that motivates the PR:
```
Meson - Platform("aarch64", "macos"; ) - clang                                      |          2      2   5.1s
Meson functional                                                                      |   14     2     16
```
with this PR: `Meson functional |   48     48  1m51.4s`.

**Real recipe** (Serd 0.32.10 from the Yggdrasil PR, *without* the `sed` workaround), `C linker for the host machine` as reported by Meson:

| target | toolchain | master | this PR |
|---|---|---|---|
| aarch64-apple-darwin | default (clang) | ❌ `Unable to detect linker` | ✅ `ld64.lld 22.1.7` |
| x86_64-apple-darwin, `preferred_gcc_version=v"8"` | default (clang) | ❌ `Unable to detect linker` | ✅ `ld64.lld 22.1.7` |
| x86_64-apple-darwin (GCC 4.8 ⇒ `clang_use_lld=false`) | default (clang) | ✅ `ld64 530` | ✅ `ld64 530` |
| x86_64-unknown-freebsd | default (clang) | ✅ `ld.bfd 2.26` | ✅ `ld.lld 22.1.7` |
| x86_64-unknown-freebsd, `preferred_gcc_version=v"8"` | default (clang) | — | ✅ `ld.lld 22.1.7` |
| x86_64-linux-gnu | default (gcc) | ✅ `ld.bfd 2.24` | ✅ `ld.bfd 2.24` |
| x86_64-w64-mingw32 | default (gcc) | — | ✅ `ld.bfd 2.24` |
| x86_64-linux-gnu (GCC 4.8 ⇒ `clang_use_lld=false`) | `_clang.meson` | ❌ `unrecognized relocation (0x2a)` | ❌ same (unchanged, bfd 2.24) |
| x86_64-linux-gnu, `preferred_gcc_version=v"8"` | `_clang.meson` | — | ✅ `ld.lld 22.1.7` |
| x86_64-w64-mingw32 (GCC 4.8 ⇒ `clang_use_lld=false`) | `_clang.meson` | ✅ `ld.bfd 2.24` | ✅ `ld.bfd 2.24` |

**Full test suite** (`Pkg.test("BinaryBuilderBase")`, default non-full-shard mode, deploy key loaded as in CI): every testset passes except `Shards/testsuite`, which fails in the Rust `cargo install` step with `failed to create directory /usr/share/testsuite/rust/cargo_build/target: Cross-device link (os error 18)`. That is https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/395 (unprivileged runner with unpacked shards; the test's skip keys off `BINARYBUILDER_RUNNER`, which was unset in my shell) and reproduces identically on unmodified master. `typos` on the diff is clean.

### Not verified

- The full-shard matrix of the new runner test (all `supported_platforms()`); only the six platforms above were run locally. CI's squashfs jobs run it.
- Docker and privileged runners.
- No Yggdrasil recipes beyond Serd were rebuilt. The Linux/FreeBSD/Windows Clang + Meson + `clang_use_lld` switch from `ld` to lld is the part worth pushing back on if you would rather keep Meson pinned to binutils there.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Claude Code 2.1.251, model: claude-fable-5)
https://claude.ai/code/session_011SdywKPCZXiDYkLvW4uCAg


src/BuildToolchains.jl
```diff
@@ -189,17 +189,19 @@ function linker_string(bt::Meson{:gcc}, p::AbstractPlatform, clang_use_lld)
...
     aatarget = aatriplet(p)
-    return clang_use_lld ? "/opt/bin/$(target)/$(lld_string(p))" : "/opt/bin/$(target)/$(aatarget)-ld"
+    return clang_use_lld ? "lld" : "/opt/bin/$(target)/$(aatarget)-ld"
 end
 
 function toolchain_file(bt::Meson, p::AbstractPlatform, envs::Dict{String,String};
                         is_host::Bool=false, clang_use_lld::Bool=false)
     target = triplet(p)
     aatarget = aatriplet(p)
-    clang_use_lld=false #Meson tries is best to misuse lld so don't use it for now
```
ChrisRackauckas, line 202: not sure about this.

ChrisRackauckas:
This isn't really in my wheelhouse but @giordano this is the upstreamed version of https://github.com/JuliaPackaging/Yggdrasil/pull/14605#discussion_r3887299826

src/BuildToolchains.jl
giordano: @gbaraldi you did this in #345, by any chance do you remember why?

gbaraldi: I don't remember exactly why. Just that it didn't work

gbaraldi:
If it passes the tests and actually uses lld in the tests then it’s fine

giordano:
@ChrisRackauckas is this ready?

ChrisRackauckas:
Yeah it all seems to be working quite well for me downstream for me.

PR #491, opened 2026-08-29
Meson: do not name a linker for Apple targets

pankgeorg:
Fixes `meson setup` failing with "Unable to detect linker" on macOS targets (e.g. JuliaPackaging/Yggdrasil#14603): the cross file names the cctools `ld64` wrapper, which Meson cannot identify, while the clang wrapper links with lld. Pointing Meson at `ld64.lld` by path does not work either (clang then drives lld without the Darwin platform/arch flags), so the Apple cross file now omits `c_ld`/`cpp_ld` and Meson uses the compiler's linker. Non-Apple cross files are unchanged.
Tested: with this package dev'd into BinaryBuilder 0.6.6, the Zix recipe (meson) builds and passes audit for `aarch64-apple-darwin` without any recipe workaround, and still for `x86_64-linux-gnu`; six Yggdrasil recipes currently carry a `sed` workaround for this that can be dropped once this is released.


giordano:
I'm inclined to favour #490 over this one

pankgeorg:
yea that makes sense. closing in favor of #490 !

PR #492, opened 2026-09-03
Rust: fix `config.toml` platform flags + add RustToolchain/RustBase 1.97.0+1

topolarity:
Publish the artifacts from https://github.com/JuliaPackaging/Yggdrasil/pull/14663 - remove the incompatible DWARF-2 exception ABI from `std`, so that no `_Unwind` symbols are emitted by `rustc` on 32-bit Windows.

Also fix an issue with our wrappers: `cargo` takes precedence over our wrappers, so most of the flags we were setting were being ignored. Synchronize these with `config.toml` so that you get the same flags when using `rustc` directly or via `cargo`. This matters on 32-bit Windows especially where we need `-C panic=abort` to dodge the exception ABI, but other platforms also had flags included.

Drafted with assistance from Claude Fable 5 🤖


giordano (approved):
🥳

PR #493, opened 2026-09-03
Rust: give cargo its own rustc wrapper via an absolute RUSTC

topolarity:
This is an alternative method for our `cargo` wrapper to inject flags to `rustc`.

As Claude 🤖 noticed in https://github.com/JuliaPackaging/Yggdrasil/pull/14679, the solution in #492 is flawed: It causes `RUSTFLAGS` provided by the user in the build script to be ignored - the `config.toml` flags override them completely.

This is an alternative fix, probably closer to what the toolchain intended to happen, which is to make sure that `cargo` actually uses our `rustc` wrapper.

A few tweaks are needed to handle `--target` correctly: `cargo` passes no target when building for the host, but our `rustc` wrapper treats the target as its default compilation target, but otherwise this works as expected and allows us to inject compilation flags.


topolarity:
I'm going to merge this since I somewhat badly broke our wrappers in #492 

@giordano let me know if you have any concerns

PR #499, opened 2026-09-11
Stop using dllist() to locate libc in uname()

maleadt:
`uname()` located libc by filtering `Libdl.dllist()` and then `dlopen`ed it, only to `dlsym` a symbol that the runtime can resolve directly. On Julia < 1.12.7, `dllist()` runs a Julia callback under glibc's dynamic linker lock, and when that callback triggers a GC while other threads are blocked on the same lock inside their own `dllist()` ccall, the process deadlocks (JuliaLang/julia#62121). BinaryBuilder's auditor constructs a `UserNSRunner` per binary from `Threads.@threads`, and every constructor runs `kernel_version_check()` -> `uname()`, so this deadlock has been killing Yggdrasil jobs several times a week.

Call `uname` through a plain ccall instead.


maleadt:
Merging this a little more quickly to fix hangs on Yggdrasil.

